PR #30, opened 2017-09-04
[jmx] Adding timeout argument to the Query function

MagdaPaj:
Cassandra integration, which uses the JMX package, fails in some cases because some of the queries takes more time than 1s. Due to this we need to have a way to configure the timeout per each query.

### Author

### Reviewer

- [x] review code for readability
- [x] verify that high risk behavior changes are well tested
- [x] check license for any new external dependency
- [x] ask questions about anything that isn't clear and obvious
- [x] approve the PR when you consider it's good to merge

### DIFF
--- a/jmx/jmx_test.go
+++ b/jmx/jmx_test.go
@@ -48,6 +48,8 @@ func TestMain(m *testing.M) {
 	}
 }
 
+const timeout = 1000
+
 func TestJmxOpen(t *testing.T) {
 	defer jmx.Close()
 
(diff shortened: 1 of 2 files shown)


jmx/jmx_test.go
```diff
@@ -36,6 +36,9 @@ func TestMain(m *testing.M) {
 			} else if command == "timeout" {
 				time.Sleep(1000 * time.Millisecond)
 				fmt.Println("{}")
+			} else if command == "noTimeout" {
```
areina: You wouldn't need this, you could use directly `timeout` command.

jmx/jmx_test.go
```diff
@@ -67,8 +70,8 @@ func TestJmxQuery(t *testing.T) {
 		t.Error()
 	}
 
-	if _, err := jmx.Query("empty"); err != nil {
-		t.Error()
+	if _, err := jmx.Query("empty", 1000); err != nil {
```
areina: You could put the `1000` value in a `const`.

MagdaPaj:
@areina I updated the code, can you have a look again?

PR #31, opened 2017-09-06
[jmx] Ignore password and username arguments if empty 

MagdaPaj:
Cassandra integration currently is failing if the customer doesn't specify password and username (so in case when JMX authentication is disable).

As a fix, in the function that composes the command sent to nrjmx tool we ignore password and username if they're empty. 

### Author

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -39,10 +39,10 @@ func getCommand(hostname, port, username, password string) []string {
 		cliCommand = []string{jmxCommand}
 	}
 
-	cliCommand = append(
-		cliCommand, "--hostname", hostname, "--port", port,
-		"--username", username, "--password", password,
-	)
+	cliCommand = append(cliCommand, "--hostname", hostname, "--port", port)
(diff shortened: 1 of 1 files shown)


jmx/jmx.go
```diff
@@ -39,10 +39,16 @@ func getCommand(hostname, port, username, password string) []string {
 		cliCommand = []string{jmxCommand}
 	}
 
-	cliCommand = append(
-		cliCommand, "--hostname", hostname, "--port", port,
-		"--username", username, "--password", password,
-	)
+	if username != "" && password != "" {
```
areina: There are some values that we want to include always (hostname & port). Then, you could simplify the code having:
```go
cliCommand = append(cliCommand, "--hostname", hostname, "--port", port)
if username != "" && password != "" {
    cliCommand = append(cliCommand, "--username", username, "--password", password)
}
```

MagdaPaj:
In the absence of Toni, @aitorciki could you have a look?

PR #33, opened 2017-12-05
Added protocol version 2 support

smoya:

This PR is a first attempt on supporting integrations protocol version 2.

### Author

### Reviewer

### DIFF
--- a/sdk/protocol2.go
+++ b/sdk/protocol2.go
@@ -71,7 +71,8 @@ func NewIntegrationProtocol2(name string, version string, arguments interface{})
 		ProtocolVersion:    "2",
 		IntegrationVersion: version,
 		prettyOutput:       defaultArgs.Pretty,
-		writer:             os.Stdout, // StdOut by default
+		Data:               []*EntityData{}, // we do because we don't want a null but an empty array on marshaling.
+		writer:             os.Stdout,       // StdOut by default
 	}
 
 	return integration, nil
@@ -96,6 +97,11 @@ func (integration *IntegrationProtocol2) Entity(entityName, entityType string) (
 			Name: entityName,
 			Type: entityType,
 		},
+
+		// empty array or object preferred instead of null on marshaling.
+		Metrics:   []metric.MetricSet{},
+		Inventory: Inventory{},
+		Events:    []Event{},
 	}
 
 	integration.Data = append(integration.Data, d)
@@ -137,7 +143,7 @@ func (integration *IntegrationProtocol2) Publish() error {
 func (integration *IntegrationProtocol2) Clear() {
 	integration.Lock()
 	defer integration.Unlock()
-	integration.Data = nil
+	integration.Data = []*EntityData{} // empty array preferred instead of null on marshaling.
 }
 
 // toJSON returns the integration as a JSON string. If the pretty attribute is
--- a/sdk/protocol2.go
+++ b/sdk/protocol2.go
@@ -28,6 +28,25 @@ type EntityData struct {
 	Events    []Event            `json:"events"`
 }
 
+// NewEntityData creates a new EntityData with default values initialised.
+func NewEntityData(entityName, entityType string) (EntityData, error) {
+	if entityName == "" || entityType == "" {
+		return EntityData{}, errors.New("entityName and entityType should be not empty")
+	}
+
+	return EntityData{
+		Entity: Entity{
+			Name: entityName,
+			Type: entityType,
+		},
+
+		// empty array or object preferred instead of null on marshaling.
+		Metrics:   []metric.MetricSet{},
+		Inventory: Inventory{},
+		Events:    []Event{},
+	}, nil
+}
+
 // IntegrationProtocol2 defines the format of the output JSON that integrations will return for protocol 2.
 type IntegrationProtocol2 struct {
 	sync.Mutex
@@ -80,10 +99,6 @@ func NewIntegrationProtocol2(name string, version string, arguments interface{})
 
 // Entity method creates or retrieves an already created EntityData.
 func (integration *IntegrationProtocol2) Entity(entityName, entityType string) (*EntityData, error) {
-	if entityName == "" || entityType == "" {
-		return nil, errors.New("entityName and entityType should be not empty")
-	}
-
 	integration.Lock()
 	defer integration.Unlock()
 	for _, e := range integration.Data {
@@ -92,21 +107,14 @@ func (integration *IntegrationProtocol2) Entity(entityName, entityType string) (
 		}
 	}
 
-	d := &EntityData{
-		Entity: Entity{
-			Name: entityName,
-			Type: entityType,
-		},
-
-		// empty array or object preferred instead of null on marshaling.
-		Metrics:   []metric.MetricSet{},
-		Inventory: Inventory{},
-		Events:    []Event{},
+	d, err := NewEntityData(entityName, entityType)
+	if err != nil {
+		return nil, err
 	}
 
-	integration.Data = append(integration.Data, d)
+	integration.Data = append(integration.Data, &d)
 
-	return d, nil
+	return &d, nil
 }
 
 // NewMetricSet returns a new instance of MetricSet with its sample attached to
--- a/sdk/protocol2.go
+++ b/sdk/protocol2.go
@@ -22,7 +22,7 @@ type Entity struct {
 
 // EntityData defines all the data related to a particular event from an entity.
 type EntityData struct {
-	Entity    Entity             `json:"entity"`
+	Entity    *Entity            `json:"entity,omitempty"` // Pointer because we want to omit the value when empty during marshaling.
 	Metrics   []metric.MetricSet `json:"metrics"`
 	Inventory Inventory          `json:"inventory"`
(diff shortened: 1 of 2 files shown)


sdk/protocol2.go
```diff
@@ -0,0 +1,173 @@
...
+// Entity is the producer of the data. Entity could be a host, a container, a pod, or whatever unit of meaning.
+type Entity struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+// EntityData defines all the data related to a particular event from an entity.
+type EntityData struct {
+	Entity    Entity             `json:"entity"`
+	Metrics   []metric.MetricSet `json:"metrics"`
```
MagdaPaj: in protocol 1 we have 
`Metrics            []*metric.MetricSet `json:"metrics"`
Why in one place we use a pointer and in the other we don't?

sdk/protocol2.go
```diff
@@ -0,0 +1,173 @@
...
+	Events    []Event            `json:"events"`
+}
+
+// EntitiesData is a collection of EntityData.
+type EntitiesData []*EntityData
+
+// Has tells you if that Entity is already part of the data.
+func (d EntitiesData) Has(entity Entity) bool {
+	for _, e := range d {
+		if e.Entity.Name == entity.Name && e.Entity.Type == entity.Type {
```
MagdaPaj: shouldn't we check that's empty also?

sdk/protocol2.go
smoya: The pointer is not needed since we are dealing with slices.

See https://github.com/newrelic/infra-integrations-beta/blob/master/integrations/kube-state-metrics/src/kube-state-metrics.go#L64

sdk/protocol2.go
smoya: We could but would not be responsibility of this method but `Entity` one.

sdk/protocol2.go
```diff
@@ -0,0 +1,173 @@
...
+
+// IntegrationProtocol2 defines the format of the output JSON that integrations will return for protocol 2.
+type IntegrationProtocol2 struct {
+	sync.Mutex
+	Name               string       `json:"name"`
+	ProtocolVersion    string       `json:"protocol_version"`
+	IntegrationVersion string       `json:"integration_version"`
+	Data               EntitiesData `json:"data"`
+	prettyOutput       bool
+	writer             io.Writer
```
MagdaPaj: What can be the use case of it?

smoya: It's being used in the `protocol_test.go`.

sdk/protocol2.go
```diff
@@ -0,0 +1,173 @@
...
+		Data:               make([]*EntityData, 0), // TODO should we make?
+		prettyOutput:       defaultArgs.Pretty,
+		writer:             os.Stdout, // StdOut by default
+	}
+
+	return integration, nil
+}
+
+// Entity method creates a new EntityData and returns it if it does not exists.
+func (integration *IntegrationProtocol2) Entity(entityName, entityType string) (*EntityData, error) {
```
MagdaPaj: I would call it `NewEntity`.

Should we have the functionality for updating `Entity`?

sdk/protocol2_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestNewIntegrationProtocol2WithDefaultArguments(t *testing.T) {
+	type argumentList struct {
+		sdk_args.DefaultArgumentList
+	}
+
+	var al argumentList
+
+	os.Args = []string{"cmd"}
+	flag.CommandLine = flag.NewFlagSet("name", 0)
```
MagdaPaj: I don't understand these 2 lines. What is their purpose? And it's not needed any check for them?

smoya: It's copy pasted from the other protocol tests. But anyway they are not needed so I'll remove them 👍

sdk/protocol2.go
smoya: The idea of this method is create or retrieve an already existing Entity. Since it's not possible to do it now (internal reasons), we could do it in the future without introducing any breaking change in the name.

sdk/protocol2.go
```diff
@@ -0,0 +1,164 @@
...
+// NewMetricSet returns a new instance of MetricSet with its sample attached to
+// the IntegrationData.
+func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
+	ms := metric.NewMetricSet(eventType)
+	d.Metrics = append(d.Metrics, ms)
+
+	return ms
+}
+
+// Publish runs all necessary tasks before publishing the data. Currently, it
```
MagdaPaj: This description is not correct with a `writer`. It should be changed to sth more general.

sdk/protocol2_test.go
```diff
@@ -0,0 +1,171 @@
...
+		t.Fail()
+	}
+
+	e2, err := i.Entity("Entity1", "test")
+	if err != nil {
+		t.Fail()
+	}
+
+	if e1 != e2 {
+		t.Error("entity should be equal.")
```
MagdaPaj: I think that this functionality is wrong. According to the requirements:

> Entity would require both a name and a type, and the combination of name+type must be globally unique.

smoya: Yes, that's why I'm checking that the Entity returned is the SAME we created before.

sdk/protocol2.go
smoya: fixed

sdk/protocol2_test.go
MagdaPaj: Yeah, you're right! but we should check in the test, that integration contains only one entity, and not 2. At least for me it's not so clear.

smoya: Ok, make sense! Done :D

sdk/protocol2.go
```diff
@@ -0,0 +1,164 @@
...
+	return integration, nil
+}
+
+// Entity method creates or retrieves an already created EntityData.
+func (integration *IntegrationProtocol2) Entity(entityName, entityType string) (*EntityData, error) {
+	if entityName == "" || entityType == "" {
+		return nil, errors.New("entityName and entityType should be not empty")
+	}
+
+	integration.Lock()
```
mariomac: I assume this process is fast enough and not too concurrent so we can afford locking the whole method until it completes.

sdk/protocol2.go
```diff
@@ -0,0 +1,164 @@
...
+		return err
+	}
+
+	output, err := integration.toJSON(integration.prettyOutput)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprint(integration.writer, output)
+	integration.Clear()
```
mariomac: if it returns error, doesn't have the integration to be cleared also?

sdk/protocol2.go
smoya: As could be as concurrent as you want we must add this lock, which in some cases could be only Read lock or Write, but trying to simplify by now.

This is a must across all our sdk. In fact we should add it in some more places.

sdk/protocol2.go
smoya: I don't think so. At this point the user should care about retrying if needed.

sdk/protocol2.go
```diff
@@ -0,0 +1,164 @@
...
+
+	integration.Data = append(integration.Data, d)
+
+	return d, nil
+}
+
+// NewMetricSet returns a new instance of MetricSet with its sample attached to
+// the IntegrationData.
+func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
+	ms := metric.NewMetricSet(eventType)
```
MagdaPaj: Setting this eventType here is problematic. Maybe the user want to initilize the metricsSet without the eventType and add it later. 
Maybe we could have one more method for `*EntityData` without setting the `eventType`

sdk/protocol2.go
```diff
@@ -0,0 +1,164 @@
...
+	// Avoid working with an uninitialized or in error state cache
+	if err = cache.Status(); err != nil {
+		return nil, err
+	}
+
+	integration := &IntegrationProtocol2{
+		Name:               name,
+		ProtocolVersion:    "2",
+		IntegrationVersion: version,
+		prettyOutput:       defaultArgs.Pretty,
```
MagdaPaj: Why don't we initialize the `Data`?
If someone doesn't populate any inventory data he will see in his final payload:
```"inventory": null```
where in the version 1 was:
```"inventory": {}``` 
(as far as I'm not wrong).

Comment (smoya) [reply sdk/protocol2.go]: done

File: sdk/protocol2.go [hunk 12]
```diff
@@ -28,6 +28,25 @@ type EntityData struct {
 	Events    []Event            `json:"events"`
 }
 
+// NewEntityData creates a new EntityData with default values initialised.
+func NewEntityData(entityName, entityType string) (EntityData, error) {
+	if entityName == "" || entityType == "" {
+		return EntityData{}, errors.New("entityName and entityType should be not empty")
```
MagdaPaj: We should remove it, not correct with requirements.

MagdaPaj:
👍

PR #36, opened 2018-01-03
Adding support for events for remote entity

MagdaPaj:
Adding support for events.

### Author

### Reviewer

### DIFF
--- a/sdk/data.go
+++ b/sdk/data.go
@@ -28,7 +28,13 @@ func (i Inventory) SetItem(key string, field string, value interface{}) {
 
 // Event is the data type to represent arbitrary, one-off messages for key
 // activities on a system.
-type Event map[string]interface{}
+type Event struct {
+	Summary  string `json:"summary"`
+	Category string `json:"category,omitempty"`
+}
+
+// DefaultEventCategory is a default value for event category
+const DefaultEventCategory = "notifications"
 
 // Integration defines the format of the output JSON that integrations will
 // return.
@@ -38,7 +44,7 @@ type Integration struct {
 	IntegrationVersion string              `json:"integration_version"`
 	Metrics            []*metric.MetricSet `json:"metrics"`
 	Inventory          Inventory           `json:"inventory"`
-	Events             []Event             `json:"events"`
+	Events             []*Event            `json:"events"`
 	prettyOutput       bool
 }
 
@@ -63,7 +69,7 @@ func NewIntegration(name string, version string, arguments interface{}) (*Integr
 		IntegrationVersion: version,
 		Inventory:          make(Inventory),
 		Metrics:            make([]*metric.MetricSet, 0),
-		Events:             make([]Event, 0),
+		Events:             make([]*Event, 0),
 		prettyOutput:       defaultArgs.Pretty,
 	}
 
@@ -76,6 +82,22 @@ func (integration *Integration) NewMetricSet(eventType string) *metric.MetricSet
 	ms := metric.NewMetricSet(eventType)
 	integration.Metrics = append(integration.Metrics, &ms)
 	return &ms
+}
+
+// AddNotificationEvent method adds a new Event with default event category.
+func (integration *Integration) AddNotificationEvent(summary string) error {
+	return integration.AddEvent(Event{Summary: summary, Category: DefaultEventCategory})
+}
+
+// AddEvent method adds a new Event.
+func (integration *Integration) AddEvent(e Event) error {
+	if e.Summary == "" {
+		return fmt.Errorf("summary of the event cannot be empty")
+	}
+
+	integration.Events = append(integration.Events, &e)
+
+	return nil
 }
 
 // Publish runs all necessary tasks before publishing the data. Currently, it
@@ -103,11 +125,11 @@ func (integration *Integration) Publish() error {
 func (integration *Integration) Clear() {
 	integration.Inventory = make(Inventory)
 	integration.Metrics = make([]*metric.MetricSet, 0)
-	integration.Events = make([]Event, 0)
+	integration.Events = make([]*Event, 0)
 }
 
 // toJSON returns the integration as a JSON string. If the pretty attribute is
-// set to true, the JSON will be idented for easy reading.
+// set to true, the JSON will be indented for easy reading.
 func (integration *Integration) toJSON(pretty bool) (string, error) {
 	var output []byte
 	var err error
--- a/sdk/protocol2.go
+++ b/sdk/protocol2.go
@@ -25,7 +25,7 @@ type EntityData struct {
 	Entity    Entity             `json:"entity"`
 	Metrics   []metric.MetricSet `json:"metrics"`
 	Inventory Inventory          `json:"inventory"`
-	Events    []Event            `json:"events"`
+	Events    []*Event           `json:"events"`
 }
 
 // NewEntityData creates a new EntityData with default values initialised.
@@ -39,7 +39,7 @@ func NewEntityData(entityName, entityType string) (EntityData, error) {
 		// empty array or object preferred instead of null on marshaling.
 		Metrics:   []metric.MetricSet{},
 		Inventory: Inventory{},
-		Events:    []Event{},
+		Events:    []*Event{},
 	}
 
 	// Entity data is optional. When not specified, data from the integration is reported for the agent's own entity.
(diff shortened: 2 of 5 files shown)


sdk/protocol2.go
```diff
@@ -132,6 +132,19 @@ func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
 	return ms
 }
 
+// Event method creates or retrieves an already created Event.
+func (d *EntityData) Event(summary, category string) *Event {
+	for _, e := range d.Events {
```
areina: I think we shouldn't check if an event is already present. Imagine you're generating an event to know if a service has been restarted: If during the interval you're monitoring events, the service has been restarted 3 times, probably you would like to have those 3 events.

sdk/protocol2.go
```diff
@@ -132,6 +132,19 @@ func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
 	return ms
 }
 
+// Event method creates or retrieves an already created Event.
+func (d *EntityData) Event(summary, category string) *Event {
```
areina: Do we need to return the reference to the event? I would say that once you have added the event to the entity data, you shouldn't work with that event anymore.

If you agree with my comments, then probably would make sense to rename the method to something like `AddEvent` to make it more explicit.

sdk/data.go
```diff
@@ -28,7 +28,19 @@ func (i Inventory) SetItem(key string, field string, value interface{}) {
...
-type Event map[string]interface{}
+type Event struct {
+	Summary  string `json:"summary"`
+	Category string `json:"category,omitempty"`
+}
+
+// NewEvent creates new Event with values passed as arguments.
+func NewEvent(summary, category string) Event {
+	ev := Event{
+		Summary:  summary,
```
areina: Which should be the behavior if I pass empty strings as `summary` or `category` arguments?

MagdaPaj: I don't remember if we discuss about it or not. Because in the current code when the category is empty we skip it in the payload. 
```
	"events": [
		{
			"summary": "foo1"
		}
]
```
But to be honest I'm not sure what should be the behaviour. I think that we should fail when `summary` is empty. And ignore `category` when is empty (as we do right now)

sdk/protocol2.go
MagdaPaj: Yeah, I was thinking about it. But you're right. We should not work with this event later. 

However if the event already exists, what should we do? Fail or do nothing? Or add the same event again? Once we decide it we can propose the name.

sdk/protocol2.go
MagdaPaj: Yes, you're right!

MagdaPaj:
Do you think we need to add more methods for events? Something similar to `SetMetric`/`SetItem` which operates only on metrics/inventory object not on integration and entity objects? But I'm not sure how easy it it. Because metric and inventory objects are simple maps, which are very easy to append.

sdk/data.go
areina: Ignoring `category` is fine because then internally the agent is assigning a default category: `notifications`. Regarding an empty summary, I agree with you: I consider it an error and we should fail.

areina:
I don't think so. Events structure is pretty simple at this moment, it's just a list of events and I consider with a `AddEvent(summary, category)` function we are fine.

sdk/protocol2.go
```diff
@@ -132,6 +132,19 @@ func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
...
+	for _, e := range d.Events {
+		if e.Summary == summary && e.Category == category {
```
smoya: I'm not sure this is the behaviour we want to adopt.

Let's say I want to log every time I restart apache. If I restart apache 5 times before flushing metrics only ` event will be reported instead of 5.

smoya: I just realised about @areina 's comment. I wrote almost the same 🤓

sdk/protocol2.go
```diff
@@ -132,6 +132,19 @@ func (d *EntityData) NewMetricSet(eventType string) metric.MetricSet {
...
+			return e
+		}
+	}
+	ev := NewEvent(summary, category)
+	d.Events = append(d.Events, &ev)
```
smoya: Have in mind that we collect event in a non thread safe way. What I mean is that if we have several goroutines calling this method at the same time will cause a race resulting in a wrong state.

There is a TODO task for our team to fix that in a general way but maybe we could do it now here.

sdk/protocol2.go
MagdaPaj: Yeah, I agree. I will remove it!

sdk/protocol2.go
areina: Good point @smoya. However, I would prefer keeping the scope of this PR, and we could address that in a different PR.

MagdaPaj: Yes, I agree with @areina, because I won't have time to finish it and test it.

MagdaPaj:
@areina, @smoya I applied requested changes and I added tests. Could you please have a look again?

sdk/data.go
```diff
@@ -28,7 +28,23 @@ func (i Inventory) SetItem(key string, field string, value interface{}) {
...
+
+// NewEvent creates new Event with values passed as arguments.
+func NewEvent(summary, category string) (Event, error) {
+	var ev Event
+	if summary == "" {
+		return ev, fmt.Errorf("summary of the event cannot be empty")
+	}
+	ev = Event{
+		Summary:  summary,
+		Category: category,
```
MagdaPaj: `category` value should be lower case always? Or we don't care about it? Because in the documentation we have examples with only lower case: https://docs.newrelic.com/docs/infrastructure/integrations-sdk/file-specifications/integration-executable-file-specifications#event-data

sdk/data.go
```diff
@@ -78,6 +94,17 @@ func (integration *Integration) NewMetricSet(eventType string) *metric.MetricSet
 	return &ms
 }
 
+// AddEvent method adds a new Event.
+func (integration *Integration) AddEvent(summary, category string) error {
```
MagdaPaj: Let's assume that user doesn't want to set `category`.
With current code he will need to do it like this:
```
integration.AddEvent("MySummary", "")
```
Maybe this would be better:
```
integration.AddEvent("MySummary")
```

sdk/data.go
smoya: I guess only lowercase. This will be done on marshalling.

MagdaPaj: @smoya What you mean that it will be done on marshalling? That even though the customer set it to uppercase then on marshalling it will be converted to lowercase?

sdk/data.go
MagdaPaj: @smoya Is there any easy way to this?

Proposal:
- changing function `AddEvent` which would accept as an argument `Event` object and validating inside which elements of `Event` are present and which not.
Example:
```
integration.AddEvent(sdk.Event{Summary:"summary", Category: "category"})
```
And then we don't have a problem when user doesn't want to include `category`. It would work like
```
integration.AddEvent(sdk.Event{Summary:"summary"})
```
Also in the future when we add more fields to the `Event` it will be easy to change `AddEvent` function.

Initially I was thinking about using variadic arguments, but then it will be hard to control it.
What do you think?

/cc @areina

sdk/data.go
smoya: I though you meant the key `category` but you said value. I guess categories should be lowercased but this is just an assumption. We must check our internals.

sdk/data.go
smoya: I would go for `integration.AddEvent("MySummary", sdk.DefaultEventCategory)`
where `DefaultEventCategory` is a const with the value `notifications`.

Also I would add the suggested categories listed on the docs: https://docs.newrelic.com/docs/infrastructure/integrations-sdk/file-specifications/integration-executable-file-specifications#event-data

MagdaPaj: I think that it won't be useful for the user having this categories list. It would be just a const right?

MagdaPaj:
@areina, @smoya Initial code refactored, please have a look when you can.
The new function usage for entity:
```
	e.AddNotificationEvent("foo")
	e.AddEvent(sdk.Event{Summary: "Foo", Category: "Bar"})
``` 
It will produce the following event output:
```
"events": [
				{
					"summary": "foo",
					"category": "notifications"
				},
				{
					"summary": "Foo",
					"category": "Bar"
				}
			]
```

sdk/data.go
MagdaPaj: There are no rules for category value. So I won't convert it to lowercase.

sdk/data.go
```diff
@@ -78,6 +84,30 @@ func (integration *Integration) NewMetricSet(eventType string) *metric.MetricSet
...
+func (integration *Integration) AddNotificationEvent(summary string) error {
+	if summary == "" {
+		return fmt.Errorf("summary of the event cannot be empty")
+	}
+	e := Event{
+		Summary:  summary,
+		Category: DefaultEventCategory,
+	}
+
+	integration.Events = append(integration.Events, &e)
```
areina: Why do you not use `AddEvent` in this function? You could simplify the code:

```go
return sdk.AddEvent(Event{Summary: summary, Category: DefaultEventCategory})
```

PR #38, opened 2018-01-05
WIP - Adding examples for tutorial code how to add events

MagdaPaj:

Adding example how to add events to the integration. 

### Author

### Reviewer

### DIFF
--- a/docs/tutorial-code/redis.go
+++ b/docs/tutorial-code/redis.go
@@ -78,9 +78,38 @@ func populateMetrics(ms *metric.MetricSet) error {
 	if err != nil {
 		return err
 	}
-
 	ms.SetMetric("net.connectionsReceivedPerSecond", metricValue, metric.RATE)
+
+	cmd = exec.Command("/bin/sh", "-c", "redis-cli info | grep uptime_in_seconds:")
+	output, err = cmd.CombinedOutput()
+	if err != nil {
+		return err
+	}
+	splittedLine = strings.Split(string(output), ":")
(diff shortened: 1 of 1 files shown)


MagdaPaj:
@mariomac, @areina This is the first commit with example of adding event. I want to have simple code but also sth real. Let me know what do you think.
 
Note: The code is getting bigger and I think it requires refactor. Changes in the code maybe won't be so big, but then it can be many things to update in the tutorial. I'm not sure what's the best moment to do it (/cc @areina)

Also we didn't include here vendor folder. I don't remember way.

docs/tutorial-code/redis.go
```diff
@@ -78,11 +78,40 @@ func populateMetrics(ms *metric.MetricSet) error {
 	if err != nil {
 		return err
 	}
-
 	ms.SetMetric("net.connectionsReceivedPerSecond", metricValue, metric.RATE)
+
+	cmd = exec.Command("/bin/sh", "-c", "redis-cli info | grep uptime_in_seconds:")
```
MagdaPaj: Now I'm thinking that I overcomplicated the code. Instead of creating a new metrics I could use an existing metric `net.connectionsReceivedPerSecond` and create an event like this:
```
err = integration.AddEvent("More than x connection received per second", "notification")
```

MagdaPaj: Code updated in the second commit.

MagdaPaj:
Close in favour of: https://github.com/newrelic/infra-integrations-sdk/pull/45

PR #42, opened 2018-01-09
Updated logrus to v1.*

smoya:

This code fixes the logrus version to `v1.*` since is the one we are using. 
This would be a BC for current users if they use an older version.

### Author

### Reviewer

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -3,7 +3,7 @@ package log
 import (
 	"os"
 
-	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


log/log.go
```diff
@@ -3,7 +3,7 @@ package log
 import (
 	"os"
 
-	"github.com/Sirupsen/logrus"
+	"github.com/sirupsen/logrus"
```
smoya: From `v1.0.0`, the package name is lowercased.

PR #46, opened 2018-01-31
Downcase import line for logrus

zlesnr:
Here we downcase import line for logrus to avoid an issue where go get tries to
include a package by the same name, but lowercased and fails.  This comes up as
part of including logrus in another project, where we want to use this SDK and
can't duplicate the logrus package with capitalizing our import.

I think the general case here is to have the package name downcased, even though
github works with either.

### Author

### Reviewer

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -3,7 +3,7 @@ package log
 import (
 	"os"
 
-	"github.com/Sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
 // SetupLogging redirects logs to stderr and configures the log level.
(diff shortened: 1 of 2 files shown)


zlesnr:
I'm unable to add a label here.

zlesnr:
I'm also unable to assign a reviewer.

MagdaPaj:
@zlesnr, thank you very much for your contribution. We're going to merge it. However we will wait with releasing it as the next version, because we will need to also update other project, which has dependency with the SDK.

PR #50, opened 2018-04-10
Improve metrics test readability

varas:
Just removing some test hieroglyphics.

### Author

### Reviewer

### DIFF
--- a/metric/metrics_test.go
+++ b/metric/metrics_test.go
@@ -4,8 +4,10 @@ import (
 	"testing"
 	"time"
 
+	"fmt"
 	"github.com/newrelic/infra-integrations-sdk/cache"
 	"github.com/newrelic/infra-integrations-sdk/metric"
+	"github.com/stretchr/testify/assert"
 )
 
 type FakeData struct {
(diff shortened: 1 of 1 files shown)


metric/metrics_test.go
```diff
@@ -4,8 +4,10 @@ import (
 	"testing"
 	"time"
 
+	"fmt"
 	"github.com/newrelic/infra-integrations-sdk/cache"
 	"github.com/newrelic/infra-integrations-sdk/metric"
+	"github.com/stretchr/testify/assert"
```
MagdaPaj: I'm not sure, but I think that it is the first place where we use `testify/assert` in SDK. I don't remember if there was any reason for NOT using it. Maybe not importing external libraries? And we don't have any `vendor` folder in SDK. This could be a problem for the customer, right?

We should be sure if we want to use it or not. Any pros/cons?
/cc: @smoya, @mariomac, @areina

mariomac: Hmmm... Maybe I understand wrong the go idiosyncrasy, but when the user runs the tests of its custom integration, the tests of our SDK won't be executed, so the user don't need to download testify or any other dependency we have only in test.

mariomac: I thought this entry in the Makefile would allow users build the project:
```Makefile
deps:
	@echo "====> Install depedencies..."
	go get -v -d -t ./...
```

varas: `stretchr/testify` is widely used and awesome, but I understand @MagdaPaj point which is nice given that we are only relaying on logrus as external dep (and imho we should also get rid of it)

Current master deps:
sdk-deps

varas:
won't add new dependencies

PR #51, opened 2018-04-10
Different packages for different protocol versions

mariomac:

sdk package divided and renamed to v1 and v2.

### Author

### Reviewer


mariomac:
I used `v1` and `v2` as packages to be concise.
Also because `sdk1` and `sdk2` could be confusing for the users (dramatisation: _"this is the SDK v3, where is the package `sdk3`"_)

sdk/v1/data.go
```diff
@@ -1,4 +1,4 @@
-package sdk
+package v1
```
varas: aren't some of these structs on v2 also?

mariomac: Same type names, different structs. The structs that are equal have been kept in `v1`, the different ones have the same name but are different, so you can instantiate them like: `v1.Integration` or `v2.Integration`.

sdk/v1/data_test.go
```diff
@@ -1,4 +1,4 @@
-package sdk_test
+package v1
```
MagdaPaj: I think that this was done on purpose. See: https://medium.com/@benbjohnson/structuring-tests-in-go-46ddee7a25c#52f8

sdk/v2/data_test.go
```diff
@@ -1,4 +1,4 @@
-package sdk_test
+package v2
```
MagdaPaj: The same as in `v1`, I would keep `v2_test`

sdk/v2/data.go
```diff
@@ -38,8 +39,8 @@ func NewEntityData(entityName, entityType string) (EntityData, error) {
 	d := EntityData{
 		// empty array or object preferred instead of null on marshaling.
 		Metrics:   []metric.MetricSet{},
-		Inventory: Inventory{},
-		Events:    []*Event{},
+		Inventory: v1.Inventory{},
```
MagdaPaj: This is confusing for me. I understand that v1 refers to protocol 1, but this is also the part of Protocol 2 and it will be always like this. This will be confusing for the users. They won't know what's coming from protocol 1, and they won't understand why they need to use sth from v1 in v2. 
I wouldn't to this change (I mean the whole PR doesn't make sense for me). But this is just my opinion.
/cc @smoya, @areina

sdk/v1/data.go
MagdaPaj: This will be confusing for users.

mariomac: Maybe I'm wrong. The way I understand go package names is that they want to provide meaning to the types they go with. That's why linters complain with constructions like: `metric.MetricSet`.

In this case, I preferred to separate types by package instead of renaming them. IMHO is cleaner having `v1.Integration` and `v2.Integration` (or any other package names) instead of `sdk.Integration` and `sdk.IntegrationProtocol2`.

sdk/v2/data.go
mariomac: I don't see it confusing, since protocol v2 uses protocol v1 as base. But If the rest of the team prefers discarding this commit, I'm ok with it.

mariomac: Also users can migrate easier to protocol v2, since they can reuse their existing types from Protocol v1.

mariomac: Anyway, in the last commit, I tried to get an intermediate solution:

```go
package v2

// Inventory is the data type for inventory data produced by an integration data
// source and emitted to the agent's inventory data store.
type Inventory v1.Inventory

// Event is the data type to represent arbitrary, one-off messages for key
// activities on a system.
type Event v1.Event
```

sdk/v1/data.go
varas: +1

sdk/v2/data_test.go
varas: why? @MagdaPaj

PR #57, opened 2018-04-13
Supress make echoes

varas:
Supress Make echoes to avoid noisy cli output.

Detailed steps should be at least optional, not mandatory, we just need to know when something goes wrong.

### Author

### Reviewer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,23 +6,19 @@ GOTOOLS = github.com/golang/lint/golint \
 all: test
 
 deps: tools
-	@echo "====> Install depedencies..."
-	go get -v -d -t ./...
+	@go get -v -d -t ./...
 
 test: deps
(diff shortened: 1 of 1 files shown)


mariomac:
Not a strong opinion here. I observed that all the makefiles of the team have echoes. Any argument to remove them?

varas:
I think detailed steps should be optional, when working we are interested on output details when something goes wrong.

That's my opinion, so feel free to close if not agree.

mariomac:
Not strong opinion. Let's see what @smoya thinks about.

smoya:
I agree removing them since the value they add is not comparable to the noise that they add when reading the Makefile.

varas:
I could add a `VERBOSE` variable to keep current verbose output in case it's required for ie: CI

PR #60, opened 2018-04-16
Decouple logger

varas:

1. Decouples logger from `logrus` via `interface` extraction and `Writer` injection.
2. Its default implementation uses Go builtin component.

A further refactor on global cache is required to remove `deprecated` `SetDebug(bool)` method.

### Author

### Reviewer

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -8,9 +8,9 @@ import (
 )
 
 type Logger interface {
-	Debug(format string, args ...interface{})
-	Info(format string, args ...interface{})
-	Error(format string, args ...interface{})
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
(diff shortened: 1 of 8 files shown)


log/log.go
```diff
@@ -3,58 +3,60 @@ package log
...
-// SetupLogging redirects logs to stderr and configures the log level.
-func SetupLogging(verbose bool) {
-	logrus.SetOutput(os.Stderr)
-	if verbose {
-		logrus.SetLevel(logrus.DebugLevel)
-	} else {
-		logrus.SetLevel(logrus.InfoLevel)
-	}
+type Logger interface {
+	Debug(format string, args ...interface{})
```
mariomac: Should we add the `-f` suffix, just to be consistent with the rest of go logging/printing functions? (Debugf, Infof, Errorf....)

varas: I also think so, but keep them without the `f` to maintain a similar previous api

PR #59, opened 2018-04-13
v2.Integration builder pattern

mariomac:
#### Description of the changes

We had NewIntegration and NewIntegrationWithWriter. I have to add a new one
and presumably the number of builder functions will continue increasing in
further refactorings (configure logging, etc...).

For this reason I have built a Builder Pattern to fluently create Integrations.

In addition, some changes have been had to be done:

- Fix GetDefaultArgs when the arguments list is a plain DefaultArgumentsList
- Allowing configure the integration to use a real mutex (Synchronized) or
  avoiding it (to remove overhead on single-threaded integrations)

### Author

- [ ] add a risk label after carefully considering the "blast radius" of your changes
- [ ] describe the _intent_ of your changes in the description. don't just rewrite your code in prose
- [ ] assign at least one reviewer

### Reviewer

- [ ] review code for readability
- [ ] verify that high risk behavior changes are well tested
- [ ] check license for any new external dependency
- [ ] ask questions about anything that isn't clear and obvious
- [ ] approve the PR when you consider it's good to merge

PR #61, opened 2018-04-16
Improved cache API

mariomac:

- Deprecated global cache
- Divided Cache in interface/implementation
- Allowed configuring cache in IntegrationBuilder

### Author

### Reviewer


cache/cache.go
```diff
@@ -21,30 +23,33 @@ func SetNow(newNow func() time.Time) {
 
 const cacheTTL = 1 * time.Minute
 
-// Cache is a map-like structure that is initialized and stored in a JSON
-// file. It also saves the timestamp when a key was stored.
-type Cache struct {
+// Cache is a key-value structure that is initialized and stored in a persistent device.
+// It also saves the timestamp when a key was stored.
+type Cache interface {
```
varas: `Cacher` would be a better Golang idiomatic name

cache/cache.go
```diff
@@ -78,9 +83,8 @@ func NewCache(l log.Logger) (*Cache, error) {
 	return cache, nil
 }
 
-// Save marshalls and stores the data a Cache is holding into disk as a JSON
-// file.
-func (cache *Cache) Save() error {
+// Save persists all the data in the cache.
+func (cache *cacheImpl) Save() error {
```
varas: how about `cache *cacheImpl` => `c *Cache` ?

Golang methods receivers tend to be 1or2 chars long

cache/cache_test.go
```diff
@@ -1,84 +1,55 @@
-package cache
+package cache_test
```
varas: same as https://github.com/newrelic/infra-integrations-sdk/pull/59#discussion_r181799985

cache/global.go
```diff
@@ -1,37 +1,82 @@
...
+const integrationsDir = "nr-integrations"
+const cacheDirFilePerm = 0755
+
+// DefaultPath returns a default folder/filename path to a cache for an integration from the given name. The name of
+// the file will be the name of the integration with the .json extension.
+func DefaultPath(integrationName string) string {
+	baseDir := filepath.Join(os.TempDir(), integrationsDir)
+	// Create integrations cache directory
+	err := os.MkdirAll(baseDir, cacheDirFilePerm)
+	if err != nil {
```
varas: Extra var `err`

cache/global.go
```diff
@@ -1,37 +1,82 @@
...
 var GlobalLog = log.NewStdErr(DefaultDebug)
 
-var instance, err = NewCache(GlobalLog)
-
 func SetupLogging(verbose bool) {
 	GlobalLog.SetDebug(verbose)
 }
 
+const integrationsDir = "nr-integrations"
+const cacheDirFilePerm = 0755
```
varas: i'd group consts before vars like:
```
const (
    constA = 1
    constB ...
)
```

cache/global.go
```diff
@@ -1,37 +1,82 @@
...
 // TODO delete when removing these globals
 var GlobalLog = log.NewStdErr(DefaultDebug)
 
-var instance, err = NewCache(GlobalLog)
-
 func SetupLogging(verbose bool) {
 	GlobalLog.SetDebug(verbose)
 }
 
+const integrationsDir = "nr-integrations"
```
varas: why this new behaviour?

cache/global.go
```diff
@@ -1,37 +1,82 @@
...
+	cachePath := os.Getenv(pathEnvVar)
+	if cachePath == "" {
+		_, fname := filepath.Split(os.Args[0])
+		cachePath = filepath.Join(os.TempDir(), fmt.Sprintf("%s-global.json", fname))
+	}
+	return cachePath
+}
+
+var instance Cache
+var err error
```
varas: grouped on top please

sdk/v2/builder.go
```diff
@@ -26,10 +27,17 @@ type IntegrationBuilder interface {
 	// Writer sets the output stream where the integration resulting payload will be written to.
 	// By default, the standard output (os.Stdout).
 	Writer(io.Writer) IntegrationBuilder
+	// Cache sets the Cache implementation that will be used to persist data between executions of the same integration.
+	// By default, it will be a Disk-backed cache named stored in the file returned by the
+	// cache.DefaultPath(integrationName) function.
+	Cache(cache.Cache) IntegrationBuilder
+	// NoCache disables the cache for this integration.
+	NoCache() IntegrationBuilder
```
varas: Not sure about this.

Integrations do need a persistent cache, I guess we should create a one per integration and hide it to the user.

Why a user would like to set it's own cache? or remove the current one?

mariomac: I think that integrations without deltas don't need a cache.

Also, they could prefer using other caches, like redis, etcd, etc....

cache/cache_test.go
mariomac: To be consistent with the previous work from the team, the reason is that this way you are "forced" to test only against public interfaces.

cache/global.go
mariomac: Just to have all the caches in a single directory, in case more than one integration coexist.

mariomac:
Good point, Juan. I'll implement it into a new patch.

PR #63, opened 2018-04-16
Enhance data serializer

varas:
Enhance data serializer by removing non required `fmt` dependency and avoiding extra byte > string > byte conversion.

Also adds lacking map `make` on Integration constructor.

### Author

### Reviewer


sdk/v2/data.go
```diff
@@ -185,7 +184,7 @@ func (integration *Integration) Clear() {
 
 // toJSON returns the integration as a JSON string. If the pretty attribute is
```
areina: super minor, but maybe we should change the comment about returning a string.

sdk/v2/data.go
```diff
@@ -183,9 +182,9 @@ func (integration *Integration) Clear() {
 	integration.Data = []*EntityData{} // empty array preferred instead of null on marshaling.
 }
 
-// toJSON returns the integration as a JSON string. If the pretty attribute is
-// set to true, the JSON will be idented for easy reading.
-func (integration *Integration) toJSON(pretty bool) (string, error) {
+// toJSON serializes integration as JSON. If the pretty attribute is
+// set to true, the JSON will be indented for easy reading.
+func (integration *Integration) toJSON(pretty bool) ([]byte, error) {
```
smoya: In a next iteration we could call this method `MarshalJSON` following the JSON Marshaler Interface from stdlib https://golang.org/pkg/encoding/json/#Marshaler

sdk/v2/data.go
```diff
@@ -196,12 +195,12 @@ func (integration *Integration) toJSON(pretty bool) (string, error) {
 	}
 
 	if err != nil {
-		return "", fmt.Errorf("error marshalling to JSON: %s", err)
+		return []byte(""), errors.Errorf("error marshalling to JSON: %s", err)
```
smoya: Do not create a new slice of byte but return `output` var directly

varas: Why? That'd change the current behaviour

PR #66, opened 2018-04-20
Removed global cache

mariomac:

This implies that the MetricsSet has to have now its own instance
of the cache. In consequence:

- the type metrics.Set is now a struct, now it has to be returned as a pointer.
- the type metrics.Set overrides now the MarshalJSON function to only marshall
  the actual metrics, according to the v2 protocol.

### Author

### Reviewer


sdk/data.go
```diff
@@ -45,13 +45,15 @@ type Event struct {
 
 // EntityData defines all the data related to a particular event from an entity.
 type EntityData struct {
-	Entity    Entity       `json:"entity"`
-	Metrics   []metric.Set `json:"metrics"`
-	Inventory Inventory    `json:"inventory"`
-	Events    []Event      `json:"events"`
+	storer    persist.Storer
```
varas: `json:"-"`

varas: otherwise it'll try to serialize storer fields https://play.golang.org/p/OBhRgNV_8FC

metric/metrics.go
```diff
@@ -24,25 +25,41 @@ const (
 )
 
 // Set is the basic structure for storing metrics.
-type Set map[string]interface{}
+type Set struct {
+	// The storer is required for deltas and rates, but forces us overriding marshaller and makes unmarshaller
```
smoya: I don't get the point about `makes unmarshaller impossible.`. The same way we implement the `MarshalJSON` it could implement `UnmarshalJSON`. I'm missing something?

metric/metrics.go
```diff
@@ -24,25 +25,41 @@ const (
...
+	// impossible. We should set back metrics. Set to a simple map, and manage it from a metrics.SetManager class, or the
+	// Integration class.
+	// This forces us, for the WHOLE integration. Do a clear separation between logic and data.
+	// TODO: global refactor, separating for all the classes, the JSON payload from the entities that manage it
```
smoya: [Minor] Go does not have classes but structs.

metric/metrics.go
```diff
@@ -24,25 +25,41 @@ const (
...
+	storer  persist.Storer
+	Metrics map[string]interface{}
+}
 
 // NewSet returns a new Set instance.
-func NewSet(eventType string) Set {
-	ms := Set{}
+// TODO: this function should not be visible to the SDK user (move it to another package?)
```
smoya: I guess this TODO is temporary. Maybe work for another PR? Could you give me a brief summary of how the user will add metric sets? Just to be aligned.

persist/storer_test.go
```diff
@@ -6,23 +6,30 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/newrelic/infra-integrations-sdk/log"
 	"github.com/newrelic/infra-integrations-sdk/persist"
 	"github.com/stretchr/testify/assert"
 )
 
+var testLog = log.NewStdErr(false)
```
smoya: Why logging to stderr instead of discarding the log?

metric/metrics.go
mariomac: We could rebuild the `Metrics` map from a stored JSON, but even if it's technically possible, I don't think it would be a good Idea to build the `Storer` instance from a stored JSON (since this configuration could change).

metric/metrics.go
mariomac: Yes, my intention was to remove it in another PR.

mariomac: And yes, My idea was to add metrics through a proxy function.

PR #67, opened 2018-04-23
Change public api

varas:
Modify public SDK API as agreed.

### Author

### Reviewer


Makefile
```diff
@@ -6,16 +6,17 @@ GOTOOLS = github.com/golang/lint/golint \
...
 deps: tools
 #	@go get -v -d -t ./...
-	@go get -v -d -t $(NODOCS) #todo:remove
+	@go get -v -d -t $(NODOCS) #TODO: remove
 
 test: deps
-#	@gocov test ./... | gocov-xml > coverage.xml
-	@gocov test $(NODOCS) | gocov-xml > coverage.xml
+	@gocov test github.com/newrelic/infra-integrations-sdk/jmx > /dev/null # TODO: fix race for jmx package
+	@gocov test -race $(PKGS) | gocov-xml >> coverage.xml
```
mariomac: Appending output to `coverage.xml`?

integration/entity.go
```diff
@@ -0,0 +1,91 @@
...
+type EntityMetadata struct {
+	Name string `json:"name"`
+	Type string `json:"type""`
+}
+
+// EntityID entity identifier
+type EntityID string
+
+// NewEntity creates a new remote-entity.
+func NewEntity(entityName, entityType string) (*Entity, error) {
```
mariomac: *NOT FOR DOING IN THIS PULL REQUEST, BUT TO HAVE IT IN MIND*

Since an `Entity` needs a `Storer`, I wonder whether it would make sense to make `Integration` as the receiver of this function, so `Integration` automatically adds the `Storer`.

Also, we will need to add synchronization to `Metrics` and `Events`, so if this function is part of `Integration`, we could add a `Lock` only if the `Integration` is synchronized.

integration/entity.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/newrelic/infra-integrations-sdk/metric"
+	"github.com/newrelic/infra-integrations-sdk/persist"
+	"github.com/pkg/errors"
+)
+
+// Entity is the producer of the data. Entity could be a host, a container, a pod, or whatever unit of meaning.
+type Entity struct {
+	lock      sync.Mutex
+	Metadata  EntityMetadata   `json:"entity"`
+	Metrics   []*metric.Set    `json:"metrics"`
```
mariomac: At the end... we decided to leave this as a pointer?

Makefile
smoya: +1 to this question. Why appending?

smoya: Do we want to test with `-race` by default? Is just a question

Makefile
```diff
@@ -6,16 +6,17 @@ GOTOOLS = github.com/golang/lint/golint \
 # Temporary patch to avoid build failing because of the outdated documentation example
 # TODO: uncomment below commented lines and remove any line that uses $(NODOCS)
 NODOCS = $(shell go list ./... | grep -v /docs/)
+PKGS = $(shell go list ./... | egrep -v "\/docs\/|jmx")
 
-all: validate test
+all: lint test
```
smoya: Do we all agree to name this target as `lint` rather than `validate`? 
If we do, I vote for changing all the rest of Makefile in our code base.

FYI, my opinion here is +1.

Makefile
varas: Yes, that'd be great, but we didn't know how to do it and it could block us for a while, we decided to keep `jmx` out of the coverage.

Feel free to update it.

integration/builder.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/newrelic/infra-integrations-sdk/args"
+	"github.com/newrelic/infra-integrations-sdk/log"
+	"github.com/newrelic/infra-integrations-sdk/persist"
+	"github.com/pkg/errors"
+)
+
+const protocolVersion = "2"
+
+// Builder OOP builder-pattern to create a new Integration instance.
+type Builder struct {
```
smoya: Rather than implementing a builder pattern here, I would vote for using functional options. Well explained in Dave Chenney's blog https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis.

`func New(name, version string, opts ...func(*Integration)`

integration/entity.go
varas: We need it like that as we are allowing the sdk user to manage the returned entity without updating the integration.

integration/entity.go
varas: Added

Makefile
varas: Yes, we want to support concurrency

Makefile
varas: I don't really mind about the name, we removed the `validate` as it was only a proxy step adding noise

integration/builder.go
smoya: Of course we could create a bunch of predefined functions for setting the Writer, etc.

varas: I like it, but I don't think it's as understandable as the builder pattern.

integration/entity.go
smoya:
```go
type Set struct {
  	storer  persist.Storer
  	Metrics map[string]interface{}
}
```
> we are allowing the sdk user to manage the returned entity

The user will be allowed to change the metrics inside the metric set no matters if `metric.Set` Metrics field is slice to pointer or not.

integration/entity.go
```diff
@@ -0,0 +1,92 @@
...
+	Inventory metric.Inventory `json:"inventory"`
+	Events    []*metric.Event  `json:"events"`
+	storer    persist.Storer
+}
+
+// EntityMetadata stores entity Metadata
+type EntityMetadata struct {
+	Name string `json:"name"`
+	Type string `json:"type""`
```
smoya: typo here (double quote is repeated)

integration/entity.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// EntityID entity identifier
+type EntityID string
+
+// NewEntity creates a new remote-entity.
+func NewEntity(entityName, entityType string, storer persist.Storer) (*Entity, error) {
```
smoya: I want to raise 2 concerns here. I will write them in 2 separate comments

integration/entity.go
```diff
@@ -0,0 +1,92 @@
...
+	// If one of the attributes is defined, both Name and Type are needed.
+	if entityName == "" && entityType != "" || entityName != "" && entityType == "" {
+		return &Entity{}, errors.New("entity name and type are required when defining one")
```
smoya: Why not returning `nil` instead of `&Entity{}` ?

integration/entity.go
```diff
@@ -0,0 +1,92 @@
...
+func (e *Entity) NewMetricSet(eventType string) (s *metric.Set, err error) {
+	s, err = metric.NewSet(eventType, e.storer)
+	if err != nil {
+		return
+	}
+
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	e.Metrics = append(e.Metrics, s)
+	return metric.NewSet(eventType, e.storer)
```
smoya: You must return `s` rather than create a new one

metric/inventory.go
```diff
@@ -0,0 +1,22 @@
...
+	if _, ok := i[key]; ok {
+		i[key][field] = value
+	} else {
+		i[key] = inventoryItem{field: value}
+	}
+
+}
+
+// NewInventory creates new inventory.
+func NewInventory() Inventory {
```
smoya: Is this method needed? I guess you added for simplifying the API but I'm not sure if it does.

Makefile
mariomac: We were referring to the `gocov-xml >> coverage.xml`, with double `>>`. Won't this generate an invalid XML?

integration/entity.go
```diff
@@ -0,0 +1,92 @@
...
+	defer e.lock.Unlock()
+	e.Events = append(e.Events, event)
+	return nil
+}
+
+// AddInventory method adds a inventory item.
+func (e *Entity) AddInventory(key string, field string, value interface{}) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	e.Inventory.SetItem(key, field, value)
```
smoya: Technically the lock should be set at Inventory level since you don't modify the `Inventory` field from the Entity but the content. Any user could call `e.Inventory.SetItem(key, field, value)` and induce a race.

integration/builder.go
areina: The builder pattern is understandable if you're familiar with it. From the point of view of a developer of an integration, I would prefer using something like:
`integration.New("redis", "2.0", integration.HTTPWriter)`
rather than:
`integration.NewBuilder("redis", "2.0").Writer(integration.HTTPWriter).Build()`

In addition to that, personally I feel `NewBuilder` confusing (we already have the concept of nr-integration-builder). 

Also, we should think that usually people will use the options that we provide, I'm not sure if they are going to write a lot of functional options by themselves. Otherwise, I don't consider it's hard to understand functional options approach.

integration/entity.go
varas: When returned in a struct maps are copied by value, I mean not the pointer value is copied.

integration/builder.go
varas: We allow the user to replace these services via fluent api:

* Logger
* ParsedArguments (not service but args)
* Synchronized (not service, just enables concurrency)
* Writer
* Storer
* InMemoryStore (replaces Storer by builtin in-memory)

You would need lots of different constructors or options to enable this flexibility. So I'd understand either:
* use a builder
* use a factory
* use functional options
* remove flexibility (services/options)

But I don't see how to easily do this adding several constructor, neither using config options (as this won't allow to customize services)

integration/entity.go
varas: I totally agree, but this involves an unplanned refactor, we may add it for the next iteration. Is that ok?

integration/entity.go
varas: nice catch!

integration/entity.go
varas: Fixed at https://github.com/newrelic/infra-integrations-sdk/pull/67/commits/e8282c623d0d4811b6f50cc4dac9afab397293e1

integration/entity.go
smoya: Sure!

integration/entity.go
smoya: Then you don't need to lock in ``AddInventory` right?

varas: We still do

PR #71, opened 2018-04-26
Code review

varas:
Code review: Default entity, rename AddInventory
+
Increase test coverage

### Author

### Reviewer


integration/entity.go
```diff
@@ -54,6 +65,11 @@ func newEntity(entityName, entityType string, storer persist.Storer) (*Entity, e
 	return &d, nil
 }
 
+// IsDefaultEntity returns true if entity is the default one (has no identifier: name & type)
```
mariomac: Do we know how the agent behaves with an entity without name nor type?

integration/entity.go
```diff
@@ -79,8 +95,8 @@ func (e *Entity) AddEvent(event *metric.Event) error {
 	return nil
 }
 
-// AddInventory method adds a inventory item.
-func (e *Entity) AddInventory(key string, field string, value interface{}) {
+// SetInventoryItem method adds a inventory item.
+func (e *Entity) SetInventoryItem(key string, field string, value interface{}) {
```
mariomac: Maybe `SetInventory` is shorter and equally meaningful (since _Item_ is very generic)

integration/integration.go
```diff
@@ -17,27 +17,41 @@ type Integration struct {
...
-	return NewBuilder(name, version).Build()
+func New(name, version string, args interface{}) (*Integration, error) {
+	return NewBuilder(name, version).ParsedArguments(args).Build()
 }
 
-// NewWithArgs creates new integration with given arguments and sane default values.
-func NewWithArgs(name, version string, args interface{}) (*Integration, error) {
-	return NewBuilder(name, version).ParsedArguments(args).Build()
+// DefaultEntity retrieves default entity to monitorize.
+func (i *Integration) DefaultEntity() *Entity {
```
mariomac: Proposal: maybe name it  `LocalEntity`?

integration/entity.go
mariomac: This public function is only used to get the default entity, I would maybe remove it or make it private. (Look to following comments for an alternative proposal).

integration/integration.go
```diff
@@ -17,27 +17,41 @@ type Integration struct {
...
+	i.locker.Lock()
+	defer i.locker.Unlock()
+
+	for _, e := range i.Entities {
+		if e.IsDefaultEntity() {
+			return e
+		}
+	}
```
mariomac: This piece of code may be a bottleneck (specially if synchronized). I would implement it the next way:

-  Add to the `Inventory` struct a private pointer to the default entity (which will not be serialized into JSON)
- If the `defaultEntity` pointer is nil, create the new default property, append it to the `i.Entities` and also point the `defaultEntity` pointer to it.
- Otherwise, return directly the `defaultEntity` pointer.

integration/entity.go
smoya: @mariomac This was discussed offline and we all agreed on set is as `SetInventoryItem` because Inventory is the whole Inventory, and what we are doing here is setting an item in the Inventory map.

integration/entity.go
varas: We should change the whole `Entitites` from list to a map as commented https://github.com/newrelic/infra-integrations-sdk/pull/71/files#diff-34fc2068cceaa93f8d6b312c44397541R53

By doing so we don't need to treat the default-entity in a separate way.

varas: Regarding the agent, good point:

We talked on this and the agent is supposed to treat that entry as no-entity case.
Btw I'll test this case and will come back here to confirm.

varas: Privatized method

varas:
Code review piggybacked on *functional options* PR https://github.com/newrelic/infra-integrations-sdk/pull/72

persist/storer.go
```diff
@@ -87,6 +89,9 @@ func NewFileStore(storePath string, l log.Logger) (Storer, error) {
 	stat, err := os.Stat(store.path)
 	// Store file doesn't exist yet
 	if err != nil {
+		if !writable(storeDir) {
+			return nil, errors.Errorf("store directory not writtable: %s", storeDir)
```
areina: typo: `writtable` -> `writable`

integration/integration.go
```diff
@@ -3,41 +3,110 @@ package integration
...
+			err = fmt.Errorf("error applying option to integration. %s", err)
+			return
+		}
+	}
+
+	// arguments
+	if err = i.checkArguments(); err != nil {
+		return
+	}
+	if err = args.SetupArgs(i.args); err != nil {
```
smoya: Now that I stop here at this line... Should not this be part of an option itself? The Setup + assignment to `i.prettyOutput` ?

integration/integration.go
smoya: We finally decided going for your proposal @mariomac. Are u going to change it here in this PR @varas ?

integration/integration_test.go
```diff
@@ -30,25 +32,55 @@ func TestBuilder_Build(t *testing.T) {
...
-	if len(i.Data) != 0 {
+	if len(i.Entities) != 0 {
 		t.Error()
 	}
 }
 
-func TestBuilder_BuildWithDefaultArguments(t *testing.T) {
+func TestDefaultIntegrationWritesToStdout(t *testing.T) {
+	// capture stdout to file
+	f, err := ioutil.TempFile("", "stdout")
```
smoya: Why do you need to write to a file? Why not creating a simple buffer?

persist/storer.go
```diff
@@ -9,6 +9,8 @@ import (
 	"time"
 
 	"github.com/newrelic/infra-integrations-sdk/log"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
```
smoya: Does `unix` package means that we will not give support to windows users?

mariomac: Good point. If there is no Go standard way to handle file permissions, we should consider creating a `storer_windows.go` in the near future.

varas: 👍 nice catch, removed

integration/integration_test.go
varas: I changed the rest of tests to a simple buffer, but this case is testing the STDOUT, so if you don't use it then you are not testing the SUT

integration/integration.go
varas: I guess so, but arguments will be changed in the near future

integration/integration.go
varas: Nope, next one https://github.com/newrelic/infra-integrations-sdk/pull/73

persist/storer.go
```diff
@@ -79,14 +82,17 @@ func NewFileStore(storePath string, l log.Logger) (Storer, error) {
...
+		if err = os.MkdirAll(storeDir, dirFilePerm); err != nil {
 			return nil, fmt.Errorf("store directory in %s could not be created", storeDir)
 		}
 	}
 
 	stat, err := os.Stat(store.path)
 	// Store file doesn't exist yet
 	if err != nil {
+		if _, err = os.OpenFile(store.path, os.O_CREATE|os.O_WRONLY, filePerm); err != nil {
+			return nil, errors.Errorf("store directory not writable: %s", storeDir)
```
smoya: Don't think this is ok. Other error could happen here, not only not writable. You must check if the eror is that running `os.IsPermission(err)`

PR #72, opened 2018-04-26
Integration creation using functional options

varas:
Integration creation using functional options.
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Code review II

### Author

### Reviewer


integration/integration.go
```diff
@@ -107,6 +162,11 @@ func (i *Integration) MarshalJSON() (output []byte, err error) {
 	return
 }
 
+// Logger returns the integration logger instance.
```
mariomac: I thought the log was provided by the user to the integration, for internal logging purposes. Do we want to expose that? In other words: when the user would need to log on behalf of the integration?

varas: I added a getter in purpose, because we are returning errors when calling: NewIntegration, Entity, NewMetricSet, Publish... and the user would probably log them on a new logger or discard them.

integration/integration.go
```diff
@@ -3,28 +3,83 @@ package integration
...
 	Name               string    `json:"name"`
 	ProtocolVersion    string    `json:"protocol_version"`
 	IntegrationVersion string    `json:"integration_version"`
 	Entities           []*Entity `json:"data"`
+	locker             sync.Locker
+	storer             persist.Storer
 	prettyOutput       bool
 	writer             io.Writer
+	logger             log.Logger
+	args               interface{} // UGLY
```
areina: If we consider this isn't good code, I would add a `TODO` note about fixing it. Please, remove the `UGLY` note.

integration/integration.go
areina: To simplify the usage, there is a default logger. The user could like to get it and use it later. What do you think?

varas: That's the current usage indeed.

PR #76, opened 2018-04-30
Rename Type for Namespace

xino12:
Rename Entity Metadata property Type to Namespace since it is more descriptive to understand later on how to search for entities in the UI.

### Author

### Reviewer


integration/entity.go
```diff
@@ -23,8 +23,8 @@ type Entity struct {
 
 // EntityMetadata stores entity Metadata
 type EntityMetadata struct {
-	Name string `json:"name"`
-	Type string `json:"type"`
+	Name      string `json:"name"`
+	Namespace string `json:"type"`
```
smoya: I would add a comment saying that, for compatibility reasons, we keep `type`

xino12: 👍

integration/entity.go
```diff
@@ -22,6 +22,7 @@ type Entity struct {
 }
 
 // EntityMetadata stores entity Metadata
+// For compatibility reasons we keep the type.
```
varas: inline this comment on the `type` line please

PR #78, opened 2018-05-02
Fix JMX race conditions

varas:
Fix JMX race conditions:
* locks on data (enables concurrency)
* cancels query context on:
  - timeout
  - empty query output

If query goroutine is not stopped, it could possible lock the integration in a `defunct` state (orphaned goroutine keeping the `nrjmx` alive).

### Author

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 var lock sync.Mutex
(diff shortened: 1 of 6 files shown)


jmx/jmx.go
```diff
@@ -128,16 +147,19 @@ func doQuery(out chan []byte, errorChan chan error, queryString []byte) {
 // Query executes JMX query against nrjmx tool waiting up to timeout (in milliseconds)
 // and returns a map with the result.
 func Query(objectPattern string, timeout int) (map[string]interface{}, error) {
+	ctx, cancelFn := context.WithCancel(context.Background())
+
 	result := make(map[string]interface{})
-	pipe := make(chan []byte)
+	lineCh := make(chan []byte, jmxLineBuffer*2)
```
varas: added buffer to avoid channel RW deadlock

jmx/jmx.go
```diff
@@ -15,16 +15,23 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/pkg/errors"
```
smoya: should be "errors" as is in stdlib

PR #83, opened 2018-05-16
Http client wrapper

xino12:
Create wrapper to ease the task of creating a http client with custom certs.

### Author

### Reviewer

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -13,14 +13,6 @@ import (
 	"time"
 )
 
-// ClientArguments are meant to be used as flags from a custom integrations. With this you could
-// send this arguments from the command line.
-type ClientArguments struct {
-	HTTPCaBundleFile string        `default: "" help: "Name of the certificate file"`
-	HTTPCaBundleDir  string        `default: "" help: "Path where the certificate exists"`
-	HTTPTimeout      time.Duration `default:30 help: "Client http timeout in seconds"`
-}
-
 // New creates a new http.Client with a custom certificate
 func New(CABundleFile, CABundleDir string, httpTimeout time.Duration) (*http.Client, error) {
 	// go default http transport settings
--- a/http/client.go
+++ b/http/client.go
@@ -1,6 +1,5 @@
-// This custom http package provides an easy way to construct an http client with custom certificates and customizable
-// timeout. If you need to customize other attributes you can use the golang http package.
-// https://golang.org/pkg/net/http/
+// Package http provides an easy way to construct an http client with custom certificates and customizable timeout.
+// If you need to customize other attributes you can use the golang http package. https://golang.org/pkg/net/http/
 package http
 
 import (
--- a/http/client.go
+++ b/http/client.go
@@ -1,3 +1,6 @@
+// This custom http package provides an easy way to construct an http client with custom certificates and customizable
+// timeout. If you need to customize other attributes you can use the golang http package.
+// https://golang.org/pkg/net/http/
 package http
 
 import (
@@ -8,10 +11,19 @@ import (
 	"net/http"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
-// NewWithCert creates a new http.Client with a custom certificate
-func NewWithCert(CABundleFile, CABundleDir string) (*http.Client, error) {
+// This arguments are meant to be used as flags from a custom integrations. With this you could easily
+// send this arguments from the command line.
+type ClientArguments struct {
+	CA_BUNDLE_FILE string `default: "" help: "Name of the certificate file"`
+	CA_BUNDLE_DIR string `default: "" help: "Path where the certificate exists"`
+	HTTP_TIMEOUT time.Duration `default:30 help: "Client http timeout"`
+}
+
+// New creates a new http.Client with a custom certificate
+func New(CABundleFile, CABundleDir string, httpTimeout time.Duration) (*http.Client, error) {
 	// go default http transport settings
 	transport := &http.Transport{}
 
@@ -24,6 +36,7 @@ func NewWithCert(CABundleFile, CABundleDir string) (*http.Client, error) {
 	}
 
 	return &http.Client{
+		Timeout:   httpTimeout * time.Second,
 		Transport: transport,
 	}, nil
 }
--- a/http/client.go
+++ b/http/client.go
@@ -3,6 +3,7 @@ package http
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


http/client.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return &http.Client{
+		Transport: transport,
+	}
+}
+
+func getCertPool(certFile string, certDirectory string) *x509.CertPool {
+	caCertPool := x509.NewCertPool()
+	var writer bytes.Buffer
+	log := log.New(false, &writer)
```
varas: inject the Logger instead of creating it here please

xino12:
@mariomac @varas Done.

http/client.go
```diff
@@ -0,0 +1,68 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
```
areina: It would be nice to have some default arguments that we could use in our integration as flags/environment variables. Ex:
```go
package main

...

type argumentList struct {
    sdkArgs.DefaultArgumentList

    http.ClientArguments
}
```
and they would expose: 
`CA_BUNDLE_FILE`, `CA_BUNDLE_DIR` and `HTTP_TIMEOUT` vars.

I think that could avoid some toil when we're developing integrations.

http/client.go
```diff
@@ -0,0 +1,68 @@
...
+	if CABundleFile != "" || CABundleDir != "" {
+		certs, err := getCertPool(CABundleFile, CABundleDir)
+		if err != nil {
+			return nil, err
+		}
+		transport.TLSClientConfig = &tls.Config{RootCAs: certs}
+	}
+
+	return &http.Client{
+		Transport: transport,
```
areina: It would be nice having a configurable timeout.

http/client.go
areina: http.ClientArguments would be something like:

```go
type ClientArguments struct {
    CA_BUNDLE_FILE string `default: "" help: "blabla"`
    CA_BUNDLE_DIR string `default: "" help: "blabla"`
    HTTP_TIMEOUT int `default:30 help:"blabla"`
}
```

http/client.go
```diff
@@ -0,0 +1,68 @@
...
+
+func getCertPool(certFile string, certDirectory string) (*x509.CertPool, error) {
+	caCertPool := x509.NewCertPool()
+	if certFile != "" {
+		caCert, err := ioutil.ReadFile(certFile)
+		if err != nil {
+			return nil, err
+		}
+
+		_ = caCertPool.AppendCertsFromPEM(caCert)
```
areina: Why are we not handling the returned value of these functions? If it isn't ok, probably we would like to debug some message, no?

http/client.go
smoya: 👍  But i would add `HTTP` prefix to the struct name

smoya:
Good job @xino12!

I took a look at the whole PR and here are some concerns i have:

* I see that there are a lot of values being customized at the Transport. I would say just modify those we already need.
* I miss some basic way to configure the main things. I think this could be a good use case for functional options as we did for the integration constructor. Some of those option could be `Cert`, `Timeout`, `Args` (passing the struct of args Toni suggested to you), etc.

http/client.go
xino12: Yes, although I try to keep this PR only about the certificates.
We can do it in another PR.

xino12:
@smoya This PR's goal is to give a way to customize only the certificates that's why we did not use functional options since that would overengineer the solution. We could do it in another PR where we improve this wrapper to allow configure other options.

smoya:
@xino12 

IMHO i find the provided code api not useful to the user mainly because it's just coupling to the only use case of setting certificates.

Also we are adding a lot of "magic" configurations such as:

```go
DialContext:           (&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second}).DialContext,
		MaxIdleConns:          100,
		IdleConnTimeout:       90 * time.Second,
		TLSHandshakeTimeout:   10 * time.Second,
		ExpectContinueTimeout: 1 * time.Second,
```

I find the functional opts a good way of decoupling things and easily configure the state of the client.

xino12:
@smoya I've deleted the "magic" options as was stated in a previous comment.
I find the functional options a good option as well, but for the goal of this PR is out of scope.(check description)
We could do it in another PR, the aim of this PR was only to provide an easy way for that use case.

http/client.go
varas: That func just returns if the cert was added to the pool

varas: Would you add a logger dep just to be able to know that in a debug?

xino12: Done

http/client.go
varas: @areina I'm not agree on using `sdkArgs.DefaultArgumentList` this should not depend on the SDK cli args, but just work as a helper within its bounded domain (https requests).

varas:
IMO we should pass **only the optional values** within a Options struct (cert stuff could be passed explicitly), this Options struct should behave with sane default values.

I'd discard the func-ops and `default` annotations approach as overkill added functionality for this tiny feature.

In the other side I don't think all the values you wrote are mean to be configurable. How many times did you change those in the past when using http clients? I'd just provide Timeout as configurable.

smoya:
I understand your point about keeping things tied and simple. I also understand the fact that you want to keep closest as possible to the PR purpose.
However IMHO the purpose of this PR should not be create a helper for generating a http client with certificates enabled, but helping the user to create http clients in a easy way, specially for those that does not have knowledge about Golang itself.

Yesterday I wrote a poc with all the suggestions made by Toni and I. I did not want to share with you in order to let us to reach a consensus before. Here the gist: https://gist.github.com/smoya/3719c207a7d50053c62e3d2f9d3550d5#file-main-go
Pay special attention to the `main` func where i'm creating the client. This would be the api, which imho is simple and powerful at the same time. Another detail is that we are not tied to any other customization so we are not tied to be the maintainers of those lines.

http/client.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+	"time"
+)
+
+// ClientArguments are meant to be used as flags from a custom integrations. With this you could
+// send this arguments from the command line.
+type ClientArguments struct {
+	HTTPCaBundleFile string        `default: "" help: "Name of the certificate file"`
+	HTTPCaBundleDir  string        `default: "" help: "Path where the certificate exists"`
+	HTTPTimeout      time.Duration `default:30 help: "Client http timeout"`
```
areina: Hey, 
have you test this? I think `time.Duration` is nanoseconds by default: https://golang.org/pkg/time/#Duration

xino12: When setting the timeout we change the units multiplying it by Time.seconds.

I'll add a description in the help.

PR #84, opened 2018-05-16
Exmplain how to declare arguments

xino12:
Improve tutorial to explain how to declare arguments.

### Author

- [] assign at least one reviewer

### Reviewer

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -658,6 +658,23 @@ type argumentList struct {
 }
 ```
 
+Arguments must have the first letter capitalised when defined in the code(ex: **H**ostname), but they cannot use any capital letter when used in the config file.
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -658,6 +658,23 @@ type argumentList struct {
 }
 ```
 
+Arguments must have the first letter capitalised when defined in the code(ex: **H**ostname), but they cannot use any capital letter when used in the config file.
+Any multi-word argument should be used in CamelCase in the code(ex: **LabelName**) and must be used as snake_case in config files(ex: **label_name**).
+In code:
+```go
+type argumentList struct {
```
mariomac: @xino12 In another PR, I would also explain that the `DefaultArgumentList` can't be used directly and has to be embedded into another structure, even if we don't add extra arguments.

PR #92, opened 2018-05-18
Using a new Storer that allows unmarshalling into structs

mariomac:

The old cache implementation didn't allow to unmarshall directly into a custom struct (only to basic types and maps).

This change breaks the previous storer API.

### Author

### Reviewer

### DIFF
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -96,17 +96,14 @@ func (ms *Set) sample(name string, value interface{}, sourceType SourceType) (fl
 
 	// Retrieve the last value and timestamp from Storer
 	var oldval float64
-	oldTime, err := ms.storer.Read(name, &oldval)
+	oldTime, err := ms.storer.Get(name, &oldval)
 	if err == persist.ErrNotFound {
 		oldval = 0
 	} else if err != nil {
 		return sampledValue, errors.Wrapf(err, "sample-key: %s", name)
 	}
 	// And replace it with the new value which we want to keep
-	newTime, err2 := ms.storer.Write(name, floatValue)
-	if err2 != nil {
-		return sampledValue, errors.Wrapf(err, "sample-key: %s", name)
-	}
+	newTime := ms.storer.Set(name, floatValue)
 
 	if err == nil {
 		duration := newTime - oldTime
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -95,17 +95,18 @@ func (ms *Set) sample(name string, value interface{}, sourceType SourceType) (fl
 	}
 
 	// Retrieve the last value and timestamp from Storer
-	oldValI, oldTime, err := ms.storer.Get(name)
 	var oldval float64
-	if err == nil {
-		oldval = oldValI.(float64)
-	} else if err == persist.ErrNotFound {
+	oldTime, err := ms.storer.Read(name, &oldval)
+	if err == persist.ErrNotFound {
 		oldval = 0
-	} else {
+	} else if err != nil {
 		return sampledValue, errors.Wrapf(err, "sample-key: %s", name)
 	}
 	// And replace it with the new value which we want to keep
-	newTime := ms.storer.Set(name, floatValue)
+	newTime, err2 := ms.storer.Write(name, floatValue)
+	if err2 != nil {
+		return sampledValue, errors.Wrapf(err, "sample-key: %s", name)
+	}
 
 	if err == nil {
 		duration := newTime - oldTime
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -73,7 +73,7 @@ func New(name, version string, opts ...Option) (i *Integration, err error) {
 
 	if i.storer == nil {
 		var err error
-		i.storer, err = persist.NewFileStore(persist.DefaultPath(i.Name), i.logger)
+		i.storer, err = persist.NewFileStore(persist.DefaultPath(i.Name), i.logger, persist.DefaultTTL)
 		if err != nil {
 			return nil, fmt.Errorf("can't create store: %s", err)
 		}
--- a/persist/storer.go
+++ b/persist/storer.go
@@ -3,6 +3,7 @@ package persist
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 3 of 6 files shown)


data/metric/metrics.go
```diff
@@ -95,17 +95,20 @@ func (ms *Set) sample(name string, value interface{}, sourceType SourceType) (fl
 	}
 
 	// Retrieve the last value and timestamp from Storer
-	oldValI, oldTime, err := ms.storer.Get(name)
-	var oldval float64
+	var oldval, oldValI float64
+	oldTime, err := ms.storer.Read(name, &oldValI)
 	if err == nil {
-		oldval = oldValI.(float64)
+		oldval = oldValI
```
varas: you can remove `oldValI`

persist/storer.go
```diff
@@ -13,158 +12,174 @@ import (
...
-	// Set sets the value under the given key, storing the current timestamp that is also returned.
-	// This method does not persist until Save is called.
-	Set(key string, value interface{}) int64
-	// Get returns the value for a given key. Last Set call timestamp is returned.
-	// If key is not found ErrNotFound error is returned.
-	Get(key string) (value interface{}, timestamp int64, err error)
-	// Delete removes the cached data for the given key
-	Delete(key string)
-	// Save persists all in-memory stored data.
-	Save() error
```
varas: not agree on removing the `Save`

integration/integration.go
```diff
@@ -127,12 +127,6 @@ func (i *Integration) Entity(name, namespace string) (e *Entity, err error) {
 // and re-initializes the integration object (allowing re-use it during the
 // execution of your code).
 func (i *Integration) Publish() error {
-	if i.storer != nil {
-		if err := i.storer.Save(); err != nil {
```
varas: why don't we flush to disk on Publish?

persist/storer.go
```diff
@@ -13,158 +12,174 @@ import (
...
-
-	err = json.Unmarshal(file, &store.inMemoryStore)
+	bytes, err := json.Marshal(&entry)
 	if err != nil {
-		l.Errorf("cannot json decode stored integration entities, starting from scratch")
+		return 0, err
 	}
-
-	return store, nil
+	return entry.Timestamp, ioutil.WriteFile(j.pathFor(key), bytes, filePerm)
```
varas: I think we shouldn't flush IO on each Write.

mariomac:
@varas you are right. Now I have submitted a new version that tries to keep a good performance (with cached unmarshalled values) and a simple interface that would allow a user loading data into any type of struct (not just floats/arrays/maps).

mariomac:
Ok. But consider that if we ever change the implementation we will need to add the error and break compatibility.

varas:
Sure @mariomac , if at anytime we consider the Storer should flush-on-write then we also need to remove the `Save` method so that'd be a BC in both interface signature and expected behaviour.

persist/storer.go
```diff
@@ -3,168 +3,222 @@ package persist
...
+// invoking a Persist() or Save() function. This will improve performance while minimizing disk usage.
 type fileStore struct {
 	inMemoryStore
 	path string
+	ilog log.Logger
 }
 
 // SetNow forces a different "current time" for the Storer.
-// This function is useful only for unit testing.
+// Deprecated: This function is useful only for unit testing outside the persist package.
```
varas: `@deprecated` is the standard docblock way

persist/storer.go
```diff
@@ -3,168 +3,222 @@ package persist
...
-	// Store file doesn't exist yet
-	if err != nil {
-		if _, err = os.OpenFile(store.path, os.O_CREATE|os.O_WRONLY, filePerm); err != nil {
-			return nil, fmt.Errorf("store directory not writable: %s", storeDir)
+		store.loadFromDisk()
+	} else if os.IsNotExist(err) {
+		folder := path.Dir(storagePath)
+		err := os.MkdirAll(folder, dirFilePerm)
+		if err != nil {
+			ilog.Errorf("can't create storage directory %q: %s", folder, err)
```
varas: This should return an error (also if returned then it should not be logged at this level)

persist/storer.go
```diff
@@ -3,168 +3,222 @@ package persist
...
 		}
-		return store, nil
+	} else {
+		ilog.Errorf("Error accessing %q: %s", storagePath, err)
```
varas: This should also return an error

persist/storer.go
mariomac: I thought the Golang's official, standard way was "Deprecated:"
https://blog.golang.org/godoc-documenting-go-code

Actually, this is the only way IntelliJ IDEA recognizes me the code as deprecated:
image

persist/storer.go
```diff
@@ -3,168 +3,222 @@ package persist
...
+			return &store, nil
 		}
-	}
 
-	stat, err := os.Stat(store.path)
-	// Store file doesn't exist yet
-	if err != nil {
-		if _, err = os.OpenFile(store.path, os.O_CREATE|os.O_WRONLY, filePerm); err != nil {
-			return nil, fmt.Errorf("store directory not writable: %s", storeDir)
+		store.loadFromDisk()
```
varas: This should return error

persist/storer.go
```diff
@@ -3,168 +3,222 @@ package persist
...
-	c.Timestamps[name] = now().Unix()
-	return c.Timestamps[name]
+func (j *fileStore) loadFromDisk() {
+	bytes, err := ioutil.ReadFile(j.path)
+	if err != nil {
+		j.ilog.Debugf("can't read %q: %s. Ignoring", j.path, err.Error())
+	} else {
+		err = json.Unmarshal(bytes, j)
+		if err != nil {
+			j.ilog.Debugf("can't unmarshall %q: %s. Ignoring", j.path, err.Error())
```
varas: return error instead of log please

persist/storer.go
varas: I was wrong! thanks for the correction!

persist/storer.go
mariomac: This could return error in two situations:
- The storage file does not exist. This would happen the first time the integration is executed, and it's not an error that should stop the integration or we should do something else but ignoring it (and log in debug mode).
- The storage file is corrupt or not marshallable. This would happen, e.g., when the integration is updated and the new data model is not compatible with the old one. I think in this case we should also just ignore the error (and log in debug mode).

persist/storer.go
mariomac: Same comment as: https://github.com/newrelic/infra-integrations-sdk/pull/92#discussion_r190528470

persist/storer.go
varas: But you should return in case you want to log it on debug mode I guess

mariomac: I think I don't catch you. The `loadFromDisk` already logs in debug ode.

mariomac: Ok, got it. You preferred log everything from the same function. Fixed.

PR #95, opened 2018-05-23
Convert Gauge to float

varas:

Convert Gauge to float

### Author

### Reviewer

### DIFF
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -68,6 +68,10 @@ func (ms *Set) SetMetric(name string, value interface{}, sourceType SourceType) 
 		if !isNumeric(value) {
 			return fmt.Errorf("non-numeric source type for gauge metric %s", name)
 		}
+		newValue, err = castToNumeric(value)
+		if err != nil {
+			return fmt.Errorf("can't convert numeric gauge into float, value: %v", value)
+		}
 	case ATTRIBUTE:
 		if _, ok := value.(string); !ok {
 			return fmt.Errorf("non-string source type for attribute %s", name)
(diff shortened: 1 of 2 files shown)


data/metric/metrics.go
```diff
@@ -85,11 +89,15 @@ func isNumeric(value interface{}) bool {
 	return err == nil
 }
 
+func castToNumeric(value interface{}) (float64, error) {
+	return strconv.ParseFloat(fmt.Sprintf("%v", value), 64)
+}
+
```
xino12: We already discussed that this solution may affect some people that are relying in some alerts to have a gauge metric wrongly reported as any type other than float. But the solution works.

data/metric/metrics.go
```diff
@@ -85,11 +89,15 @@ func isNumeric(value interface{}) bool {
 	return err == nil
 }
 
+func castToNumeric(value interface{}) (float64, error) {
```
MagdaPaj: I have just one picky comment, the rest is fine :) 
Having this function `castToNumeric`, is it still needed to have `isNumeric` https://github.com/newrelic/infra-integrations-sdk/pull/95/files#diff-83416e4d39b0688a4bec5f4a2c076486R87? Shouldn't we just get rid of `isNumeric`?

data/metric/metrics.go
varas: True

I don't agree with the whole metrics api, imho we should enforce types by splitting the `SetMetric(name string, value interface{}, sourceType SourceType)` into different methods with different signatures like `SetGauge(name string, value float)`.

This change is a clear point where the user could expect some troubles.

Wdyt? @areina @MagdaPaj

Issue #48, opened 2018-04-05
Cache collisions for metric.RATE metrics


bryangwilliam:
I am getting the following error from the sdk for the custom integration I am writing which uses the `metric.RATE` option:
```
WARN[0000] Error setting value: Samples for queue.jobsPerSecond are too close in time, skipping sampling 
WARN[0000] Error setting value: Samples for queue.jobsPerSecond are too close in time, skipping sampling 
WARN[0000] Error setting value: Samples for queue.jobsPerSecond are too close in time, skipping sampling 
```
The above example came from trying to add a metric named `queue.jobsPerSecond` to 4 different MetricSets. The problem seems to stem from the fact that names are global in the cache rather than being unique per MetricSet, which means that MetricSets which use the same metric name will have a collision.

Adding a unique prefix per MetricSet to the key would get rid of the problem.

##### Integrations SDK version

- [x] 2.0

##### Golang version

- [x] All of them

##### OS

- [x] All of them

matiasburni:
Hi @bryangwilliam, we're going to take a look.

varas:
Hi @bryangwilliam this issue should be fixed in the oncoming SDKv3 update  https://github.com/newrelic/infra-integrations-sdk/pull/106

So you can use `NewMetricSet(eventType string, nameSpacingAttributes ...metric.Attribute)` the optional list of attributes will namespace the metric, so they collision is avoided. See https://github.com/newrelic/infra-integrations-sdk/pull/106/files#diff-2fa204c6e91bc2d1cefbda792b92f384R82 

A usage sample could be: `NewSet("type", storeWrite, Attr("pod", "pod-a"))`

PR #106, opened 2018-06-06
Cache collision fix

varas:

Fixes cache collision on the store for _rate&deltas_ when a **same metric name** is used.

Attributes are supported to namespace the metric name on the _NewSet_ constructor.

Fix test at `TestNewSet_Attr_SolvesCacheCollision` https://github.com/newrelic/infra-integrations-sdk/pull/106/files#diff-d41479d78f43948797a5970be862f5e0R196

### Author

### Reviewer

### DIFF
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -14,8 +14,8 @@ import (
 // each one.
 type SourceType int
 
-// KV represents a metric key-value pair.
-type KV struct {
+// Attribute represents an attribute metric in key-value pair format.
+type Attribute struct {
 	Key   string
 	Value string
 }
@@ -41,27 +41,38 @@ var (
 
 // Set is the basic structure for storing metrics.
 type Set struct {
-	storer  persist.Storer
-	Metrics map[string]interface{}
-	id      []KV
+	storer    persist.Storer
+	Metrics   map[string]interface{}
+	relatedTo []Attribute
 }
 
-// NewSet creates new metrics set, optionally identified by a list of key-values.
-func NewSet(eventType string, storer persist.Storer, kv ...KV) (*Set, error) {
-	ms := Set{
-		Metrics: make(map[string]interface{}),
-		storer:  storer,
-		id:      kv,
+// NewSet creates new metrics set, optionally related to a list of attributes.
+// If related attributes are used, then a new attribute-metric is added per kv-pair.
+func NewSet(eventType string, storer persist.Storer, relatedTo ...Attribute) (s *Set, err error) {
+	s = &Set{
+		Metrics:   make(map[string]interface{}),
+		storer:    storer,
+		relatedTo: relatedTo,
 	}
 
-	err := ms.SetMetric("event_type", eventType, ATTRIBUTE)
+	err = s.SetMetric("event_type", eventType, ATTRIBUTE)
+	if err != nil {
+		return
+	}
 
-	return &ms, err
+	for _, attr := range relatedTo {
+		err = s.SetMetric(attr.Key, attr.Value, ATTRIBUTE)
+		if err != nil {
+			return
+		}
+	}
+
+	return
 }
 
-// NewKV creates a new KV pair.
-func NewKV(key string, value string) KV {
-	return KV{
+// RelatedTo creates an attribute for a metric-set to be related to.
+func RelatedTo(key string, value string) Attribute {
+	return Attribute{
 		Key:   key,
 		Value: value,
 	}
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -14,10 +14,10 @@ import (
 // each one.
 type SourceType int
 
-// group identifies a metric group
-type group struct {
-	field string
-	value string
+// KV represents a metric key-value pair.
+type KV struct {
+	Key   string
+	Value string
 }
 
 const (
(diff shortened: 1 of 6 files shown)


data/metric/metrics.go
```diff
@@ -82,44 +116,98 @@ func (ms *Set) SetMetric(name string, value interface{}, sourceType SourceType)
...
 	return json.Marshal(ms.Metrics)
 }
+
+// Required for Go < v.18, as these do not include sort.Slice
+
+// Attributes list of attributes
+type Attributes []Attribute
+
+// Len ...
+func (a Attributes) Len() int { return len(a) }
```
mariomac: What are `Len`, `Swap` and `Less` used for?

varas: it's on the comment above: https://github.com/newrelic/infra-integrations-sdk/pull/106/files#diff-83416e4d39b0688a4bec5f4a2c076486R196

> Required for Go < v.18, as these do not include sort.Slice

you can see that change isolated in the commit https://github.com/newrelic/infra-integrations-sdk/pull/106/commits/7da060855a4e41e7ee36a87f451c25ee18e59229

mariomac: Hmm ok. I forgot the implicit implementation of interfaces in Go.

data/metric/metrics.go
```diff
@@ -25,48 +32,75 @@ const (
 	ATTRIBUTE SourceType = iota
 )
 
+const (
+	// NSSeparator is the metric namespace separator
+	NSSeparator = "::"
+	// NSAttributeSeparator is the metric attribute key-value separator applied to generate the metric ns.
+	NSAttributeSeparator = "=="
```
smoya: Shall we use `|` just to keep consistency with our entity GUID spec?

data/metric/metrics.go
```diff
@@ -25,48 +32,75 @@ const (
...
+)
+
+// Errors
+var (
+	ErrNonNumeric        = errors.New("non-numeric value for rate/delta")
+	ErrNoStoreToCalcDiff = errors.New("can't use deltas nor rates without persistent store")
```
smoya: Avoid using contractions on error messages. You can use `cannot` instead.

data/metric/metrics.go
```diff
@@ -14,6 +15,12 @@ import (
 // each one.
 type SourceType int
 
+// Attribute represents an attribute metric in key-value pair format.
+type Attribute struct {
```
smoya: Why not declaring Attribute as `map[string]string` ? Or why not using directly a map as users will not need to use that custom type?

data/metric/metrics.go
varas: what's the gain on switching `ErrNoStoreToCalcDiff`to `ErrCannotCalcDiffWithoutStore`?

data/metric/metrics.go
smoya: Oh I see. You want to keep the previous interface with no attributes specified.

smoya: If we don't enforce users to set those values, they will have eventually the same issue we are trying to fix here.

varas: The point on types (besides safety) is expressing intention. So you understand what that data is and what is aimed to do.

data/metric/metrics.go
smoya: I'm talking about the error message.

data/metric/metrics.go
varas: This is just a disk-storage implementation detail. Nothing related to be managed by user, neither to be pushed to agent. So entity naming here would be indeed misleading.

smoya: 👍

varas: What I'd do is privatize these consts

data/metric/metrics.go
varas: ops, sure 👍

data/metric/metrics.go
```diff
@@ -82,44 +116,98 @@ func (ms *Set) SetMetric(name string, value interface{}, sourceType SourceType)
 	return nil
 }
 
-func castToNumeric(value interface{}) (float64, error) {
+func castToFloat(value interface{}) (float64, error) {
```
smoya: 👍

data/metric/metrics.go
varas: The point on enforcing that was discussed with @areina.

We agreed that there is no need for that as most of the users won't need the feature, so enforcing them to solve a problem they don't have is adding undesired complexity.

smoya: IMHO we should enforce it since the user is not aware of the real problem when using RATE and DELTA: they may be facing the issue without even being aware of it.

smoya: The big issue here, and what I consider we must fix asap is How to friendly let the user know that the metric sets should be unique.

varas: Added error as a warning (does not breaks the logic) when SetMetric on rate/gauge in a metric-set without attributes

data/metric/metrics.go
```diff
@@ -14,112 +15,212 @@ import (
...
 }
 
-// NewSet creates new metrics set.
-func NewSet(eventType string, storer persist.Storer) (*Set, error) {
-	ms := Set{
-		Metrics: map[string]interface{}{},
-		storer:  storer,
+// NewSet creates new metrics set, optionally related to a list of attributes. These attributes makes the metric-set unique.
+// If related attributes are used, then new attributes are added.
+// TODO remove obsolete returned error
```
varas: This is a BC from previous beta, but really nice to have as it removes a BC from SDKv2 =)

data/metric/metrics.go
varas: https://github.com/newrelic/infra-integrations-sdk/pull/106/files#diff-83416e4d39b0688a4bec5f4a2c076486R98

smoya:
This PR fixes https://github.com/newrelic/infra-integrations-sdk/issues/48

PR #109, opened 2018-06-20
Fixed nil pointer when creating an entity

davidbrota:
Inside Entity creation there is a loop that checks previous entities' Metadata (Name and Namespace). The problem is that local entities don't have an associate Metadata, so if you create a local entity and then a remote entity a nil pointer crash happens.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -252,6 +252,18 @@ func newTestIntegration(t *testing.T) *Integration {
 	return i
 }
 
+func TestIntegration_CreateLocalAndRemoteEntities(t *testing.T) {
+	i, err := New(integrationName, integrationVersion)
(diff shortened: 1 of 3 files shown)


integration/integration_test.go
```diff
@@ -252,6 +252,18 @@ func newTestIntegration(t *testing.T) *Integration {
 	return i
 }
 
+func TestIntegration_CreateLocalAndRemoteEntities(t *testing.T) {
+	i, err := New(integrationName, integrationVersion)
+	assert.NoError(t, err)
+
+	local := i.LocalEntity()
+	assert.NotEqual(t, local, nil)
```
smoya: Minor: `assert.NotNil` does the job

PR #111, opened 2018-06-26
Remove non-concurrency support

xino12:

Remove non-concurrency support since we found that there was no bottleneck.

### Author

### Reviewer


docs/toolset/integration.md
```diff
@@ -23,30 +23,29 @@ The above invocation will return an integration with the default configuration:
...
         integration.Logger(logrus.New()),
-        integration.Synchronized(),
         integration.Writer(payloadFile),
     )
 ```
 
 For more details, check the documentation of the
 [Option interface implementations](https://godoc.org/github.com/newrelic/infra-integrations-sdk/integration#Option).
 
+You can safely add data from different concurrent threads since the sdk is thread safe.
```
mariomac: Can we be 100% sure to do this statement?

xino12: Not sure if 100%, maybe I could rephrase that?

AFAIK tests are running with -race flag that I understood is testing if there could be problems of concurrency and we have Test with goroutines manipulating data structures. But can't be sure 100%.

PR #112, opened 2018-06-27
Cli-arguments override Env vars

xino12:
Override cli-arguments over Env var.

### Author

### Reviewer


args/args_test.go
```diff
@@ -126,7 +126,7 @@ func TestEnvironmentVarsOverrideCliArgs(t *testing.T) {
 	clearFlagSet()
 	assert.NoError(t, sdk_args.SetupArgs(&args))
 
-	assert.False(t, args.Verbose)
+	assert.True(t, args.Verbose)
```
mariomac: The name of the test should change to `TestCliArgsOverrideEnvironmentArgs`.
As "meta-test" I would also set an environment flag different than the default one, without CLI, and see that in this case, the Environment var has been successfully used.

xino12: So true, I've changed the test name, regarding the environment vars there is another test already in:  https://github.com/newrelic/infra-integrations-sdk/blob/master/args/args_test.go#L84

PR #116, opened 2018-07-04
Tutorial update

xino12:
### Author

### Reviewer


docs/tutorial.md
```diff
@@ -1,13 +1,19 @@
 # Tutorial
 
+## Important note:
```
mariomac: I would remove "In this new version we've added". I would also remove this section and move these links to [The point where we create the integration and we choose a local entity](https://github.com/newrelic/infra-integrations-sdk/pull/116/files#diff-0adced32de34cd29581087ebf0482058R137), briefly explaining why it has sense to choose a local entity.

docs/tutorial.md
```diff
@@ -492,16 +516,16 @@ If you do not see your metric data in New Relic Insights, please check [configur
...
-	return nil
+// Add Inventory item
+if args.All() || args.Inventory {
+    // Insert here the logic of your integration to get the inventory data
+	// err = entity.SetInventoryItem("instance", "version", "3.0.1")
+    //panicOnErr(err)
 }
 ```
-Notice that in the code above we use the `Inventory` type from the SDK package for inventory data. The `SetItem` method stores a value into the inventory data structure. The first argument is the name of the inventory item, and the other two are a field name and the inventory data value.
+Notice that in the code above we use the `SetInventoryItem` method stores a value into the inventory data structure. The first argument is the name of the inventory item, and the other two are a field name and the inventory data value.
```
alejandrodnm: Replacing the "*stores a value*" with "*to store a value*" makes the sentence more clear.  

> Notice that in the code above we use the `SetInventoryItem` method to store a value into the inventory data structure

docs/tutorial.md
```diff
@@ -746,54 +777,77 @@ you will receive (value will vary):
 uptime_in_seconds:54782
 ```
 
-We assume that when the uptime is less than 60 seconds, the Redis service has recently started. Let's create a new function called `populateEvents`. It will call the `redis-cli info | grep uptime_in_seconds:` command and then create a notification event if the uptime value is smaller than a defined limit. To do so, we will use the method `AddNotificationEvent`, which adds a new event with the default `notifications` category for an integration object. It accepts the `string` argument, which is a summary message, i.e. `"Redis Server recently started"`.
+We assume that when the uptime is less than 60 seconds, the Redis service has recently started. We will call the `redis-cli info | grep uptime_in_seconds:` command and then create a notification event if the uptime value is smaller than a defined limit. To do so, we will use the type of event `event.NewNotification`, which is a event with the default `notifications` category for an integration object. It accepts the `string` argument, which is a summary message, i.e. `"Redis Server recently started"`.
```
alejandrodnm: > which is a event

Should be 

> which is an event

docs/tutorial.md
```diff
@@ -746,54 +777,77 @@ you will receive (value will vary):
...
+	// Add Event
+	if args.All() || args.Events {
+		uptime, err := queryRedisInfo("uptime_in_seconds:")
+		panicOnErr(err)
+		if uptime < 60 {
+			err = entity.AddEvent(event.NewNotification("Redis Server recently started"))
+		}
+		panicOnErr(err)
+		if uptime < 60 {
+			err = entity.AddEvent(event.New("Redis Server recently started", "redis-server"))
```
alejandrodnm: This `Event` doesn't belong in this example, you're adding it in the next one to show how to specify the category. If you look at the resulting json below, the event is not there.

xino12: nice catch!

docs/tutorial.md
xino12: I've rephrased it but I think this section is important to avoid people looking for multiple entities go through all the tutorial which is not short.

docs/tutorial-code/redis.go
```diff
@@ -25,124 +25,82 @@ const (
...
 }
 
 func main() {
+	// Create Integration
 	i, err := integration.New(integrationName, integrationVersion, integration.Args(&args))
 	panicOnErr(err)
 
-	e := i.LocalEntity()
+	// Create Entity, entities name must be unique
+	entity := i.LocalEntity()
```
davidbrota: Is the message above correct? I think there is only one local entity and it has no name. Maybe I'm wrong though

docs/tutorial-code/single-entity/redis.go
```diff
@@ -0,0 +1,105 @@
...
+	sdkArgs "github.com/newrelic/infra-integrations-sdk/args"
+	"github.com/newrelic/infra-integrations-sdk/data/event"
+	"github.com/newrelic/infra-integrations-sdk/data/metric"
+	"github.com/newrelic/infra-integrations-sdk/integration"
+)
+
+type argumentList struct {
+	sdkArgs.DefaultArgumentList
+	Hostname string `default:"localhost" help:"Hostname or IP where Redis server is running."`
+	Port     int    `default:"6379" help:"Port on which Redis server is listening."`
```
davidbrota: The Hostname and Port arguments are being used?

docs/tutorial.md
```diff
@@ -165,103 +182,88 @@ _integration_version_),
...
-	fatalIfErr(err)
-
-	// the code for populating Inventory and Metrics omitted
-
-	fatalIfErr(integration.Publish())
+    i, err := integration.New(integrationName, integrationVersion)
+    fatalIfErr(err)
+    
+    // Create Entity, entities name must be unique
+    entity := i.LocalEntity()
```
davidbrota: Misleading comment on local entity

docs/tutorial.md
```diff
@@ -282,38 +284,53 @@ source type in these cases. For metric names, it is recommended that you use a p
...
+	return strconv.ParseFloat(strings.TrimSpace(splittedLine[1]), 64)
+}
 
-	return nil
+func main() {
+    i, err := integration.New(integrationName, integrationVersion)
+    panicOnErr(err)
+    
+    // Create Entity, entities name must be unique
+    entity := i.LocalEntity()
```
davidbrota: Misleading comment on local entity

PR #127, opened 2018-07-18
Doc for RATE & DELTA required attribute/s

varas:
Update doc to target RATE & DELTA required attribute/s

### Author

### Reviewer

### DIFF
--- a/docs/toolset/integration.md
+++ b/docs/toolset/integration.md
@@ -95,11 +95,18 @@ myHost := payload.LocalEntity()
 
 cpu, err := myHost.NewMetricSet("CpuSample")
 cpu.SetMetric("cpuPercent", 75.0, metric.GAUGE)
+```
 
-disk, err := myHost.NewMetricSet("DiskSample")
+RATE and DELTAs requires to belong to at least 1 attribute. As they are persisted on disk this attribute is used to "namespace" the metrics on the set so they don't collide with others with the same name.
+
+So `NewMetricSet` provides an optional list of `metric.Attribute` arguments. There's a constructor function that comes handy to create one: `metric.Attr`.
+
+If no `Attribute` is provided, an `error` value will be returned when adding a RATE or DELTA to the metric-set.
+
+```go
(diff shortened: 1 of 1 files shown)


docs/toolset/integration.md
```diff
@@ -95,11 +95,18 @@ myHost := payload.LocalEntity()
...
 cpu, err := myHost.NewMetricSet("CpuSample")
 cpu.SetMetric("cpuPercent", 75.0, metric.GAUGE)
+```
+
+RATE and DELTAs requires to belong to at least 1 attribute. As they are persisted on disk this attribute is used to "namespace" the metrics on the set so they don't collide with others with the same name.
+
+So `NewMetricSet` provides an optional list of `metric.Attribute` arguments. There's a constructor function that comes handy to create one: `metric.Attr`.
 
-disk, err := myHost.NewMetricSet("DiskSample")
+If no `Attribute` is provided, an `error` value will be returned when adding a RATE or DELTA to the metric-set.
```
alejandrodnm: I would specify that the `Attribute` should be provided to the `NewMetricSet` at the time of declaring it, something like

> If no `Attribute` is provided to the new`NewMetricSet` function, an `error` value will be returned when adding a RATE or DELTA to the metric-set.

Just to remove any kind of confusion with adding the `Attribute` after declaring the metric-set with the `SetMetric` function

docs/toolset/integration.md
```diff
@@ -95,11 +95,18 @@ myHost := payload.LocalEntity()
 
 cpu, err := myHost.NewMetricSet("CpuSample")
 cpu.SetMetric("cpuPercent", 75.0, metric.GAUGE)
+```
+
+RATE and DELTAs requires to belong to at least 1 attribute. As they are persisted on disk this attribute is used to "namespace" the metrics on the set so they don't collide with others with the same name.
```
alejandrodnm: The `s` in `requires` is not necessary 
 
> RATE and DELTAs requires

I think that you can change the phrase to:

> RATEs and DELTAs should be grouped with at least 1 attribute

What do you think?

varas: I think the last sentence is misleading 🤔

PR #143, opened 2018-09-27
add custom attributes

xino12:
Add a way using environment variables with a specific prefix to decorate samples with custom attributes.

### Author

### Reviewer

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -14,11 +14,12 @@ import (
 // DefaultArgumentList includes the minimal set of necessary arguments for an integration.
 // If all data flags (Inventory, Metrics and Events) are false, all of them are published.
 type DefaultArgumentList struct {
-	Verbose   bool `default:"false" help:"Print more information to logs."`
-	Pretty    bool `default:"false" help:"Print pretty formatted JSON."`
-	Metrics   bool `default:"false" help:"Publish metrics data."`
-	Inventory bool `default:"false" help:"Publish inventory data."`
-	Events    bool `default:"false" help:"Publish events data."`
+	Verbose          bool `default:"false" help:"Print more information to logs."`
+	Pretty           bool `default:"false" help:"Print pretty formatted JSON."`
+	Metrics          bool `default:"false" help:"Publish metrics data."`
+	Inventory        bool `default:"false" help:"Publish inventory data."`
+	Events           bool `default:"false" help:"Publish events data."`
+	CustomAttributes bool `default:"false" help:"Decorate Samples with Custom Attributes."`
 }
 
 // All returns if all data should be published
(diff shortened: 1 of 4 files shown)


integration/entity.go
```diff
@@ -18,6 +18,8 @@ type Entity struct {
 	Events    []*event.Event       `json:"events"`
 	storer    persist.Storer
 	lock      sync.Locker
+	// CustomAttributes []metric.Attribute `json:"custom_attributes,omitempty"`
+	customAttributes []metric.Attribute
```
xino12: This here is just commented because we could do it either by adding the custom attributes to the samples in the integration. Or add them to the entities and then in our backend decorate all the samples from that entity. That's why I've left this here to show how we could do it.

integration/entity.go
```diff
@@ -72,8 +74,13 @@ func (e *Entity) isLocalEntity() bool {
 
 // NewMetricSet returns a new instance of Set with its sample attached to the integration.
 func (e *Entity) NewMetricSet(eventType string, nameSpacingAttributes ...metric.Attribute) *metric.Set {
+
 	s := metric.NewSet(eventType, e.storer, nameSpacingAttributes...)
 
+	if len(e.customAttributes) > 0 {
+		metric.AddCustomAttributes(s, e.customAttributes)
+	}
+
```
xino12: Won't be necessary if we do it in the backend.

integration/entity.go
```diff
@@ -98,3 +105,8 @@ func (e *Entity) SetInventoryItem(key string, field string, value interface{}) e
 	defer e.lock.Unlock()
 	return e.Inventory.SetItem(key, field, value)
 }
+
+func (e *Entity) setCustomAttribute(key string, value string) {
+	attribute := metric.Attribute{key, value}
+	e.customAttributes = append(e.customAttributes, attribute)
+}
```
xino12: Won't be necessary if we do it in the backend.

args/args.go
```diff
@@ -14,11 +14,12 @@ import (
...
-	Pretty    bool `default:"false" help:"Print pretty formatted JSON."`
-	Metrics   bool `default:"false" help:"Publish metrics data."`
-	Inventory bool `default:"false" help:"Publish inventory data."`
-	Events    bool `default:"false" help:"Publish events data."`
+	Verbose          bool `default:"false" help:"Print more information to logs."`
+	Pretty           bool `default:"false" help:"Print pretty formatted JSON."`
+	Metrics          bool `default:"false" help:"Publish metrics data."`
+	Inventory        bool `default:"false" help:"Publish inventory data."`
+	Events           bool `default:"false" help:"Publish events data."`
+	CustomAttributes bool `default:"false" help:"Decorate Samples with Custom Attributes."`
```
chaupt: This makes sense to start by default with `false`

varas: I'd use a non-composed name here as `CustomAttributes` would become `-custom_attributes` what's kind weird to use snake case on cli args

varas: Ie: https://github.com/newrelic/infra-integrations-sdk/pull/144

PR #150, opened 2018-11-07
Entities require name and type

varas:

Entities require name and type, if empty an error should be returned. Otherwise we'll be using a local-entity under the hood.

### Author

### Reviewer


varas:
@mariomac  there's no need, @davidbrota identified an unexpected behaviour being: when no name & type is provided to remote-entity there's no error shown but behind the scene the infra agent will treat this "remote-entity" as "local-entity"

PR #151, opened 2018-11-13
Update integration.Publish method to print breaklines

areina:

The contract between the agent and integrations is that integrations
would print a single line output. If the execution of the integration
binary returns multiple integration outputs, they have to be separated
by a break line.

You can compare the current behavior with v2.2.0:

https://github.com/newrelic/infra-integrations-sdk/blob/v2.2.0/sdk/data.go#L95

As you can see, in that version we were using `Println()`

### Author

### Reviewer

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -61,7 +61,7 @@ func TestDefaultIntegrationWritesToStdout(t *testing.T) {
 	f.Close()
 	payload, err := ioutil.ReadFile(f.Name())
 	assert.NoError(t, err)
-	assert.Equal(t, `{"name":"integration","protocol_version":"2","integration_version":"4.0","data":[]}`, string(payload))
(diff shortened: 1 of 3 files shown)


integration/integration_test.go
```diff
@@ -61,7 +61,7 @@ func TestDefaultIntegrationWritesToStdout(t *testing.T) {
 	f.Close()
 	payload, err := ioutil.ReadFile(f.Name())
 	assert.NoError(t, err)
-	assert.Equal(t, `{"name":"integration","protocol_version":"2","integration_version":"4.0","data":[]}`, string(payload))
+	assert.Equal(t, "{\"name\":\"integration\",\"protocol_version\":\"2\",\"integration_version\":\"4.0\",\"data\":[]}\n", string(payload))
```
varas: as it's just a test, to concat with `+ "\n"` would be more readable

varas: wdyt? https://github.com/newrelic/infra-integrations-sdk/pull/152

areina: I agree

PR #166, opened 2019-02-12
Refactor: remove duplication on JMX

varas:

Refactor: remove duplication on JMX because of SSL

### Author

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -39,97 +39,34 @@ const (
 	jmxLineBuffer = 4 * 1024 * 1024 // Max 4MB per line. If single lines are outputting more JSON than that, we likely need smaller-scoped JMX queries
 )
 
-func getCommand(hostname, port, username, password string) []string {
-	var cliCommand []string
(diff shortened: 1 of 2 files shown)


jmx/jmx.go
```diff
@@ -39,97 +39,34 @@ const (
...
-
-		lock.Lock()
-		defer lock.Unlock()
-		cmd = nil
-
-		done.Done()
-	}()
-
-	return nil
+	return OpenWithSSL(hostname, port, username, password, "", "", "", "")
```
mariomac: Methods code are not 100% equal. `Open` invoked `getCommand` while `OpenWithSSL` invokes `getCommandWithSSL`... won't this change break non-ssl connections?

varas: If you look at `getCommandWithSSL` it does the same as `getCommand` but only appends ssl args when args are not empty, so calling `Open` is the same as calling always `OpenWithSSL` with empty ssl args

PR #167, opened 2019-02-20
Make cluster and service entity attributes

alejandrodnm:

Make the cluster and service entity attributes instead of setting them as custom attributes.

This change will depend on the agent doing the cluster and service decoration. Integrations made with this version of the SDK should have a clear version dependency with the agent version that implements this feature, to not break backwards compatibility, meaning stop decorating metrics with this values.

Also review the documentation draft. https://docs.newrelic.com/node/12561/draft

### Author

### Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - SSL support for jmx package. Explained at https://github.com/newrelic/infra-integrations-sdk#jmx-support
 
+## Next Release
+
+### Changed
+
+- Protocol v2.1. NriCluster and NriService are set as entity attributes,
+  instead of decorating every sample.
+
 ## 3.1.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Next Release
+
+### Added
+
+- Protocol v3. NriCluster and NriService are set as entity metadata fields,
+  instead of entity custom attributes.
+
 ## 3.1.1
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


integration/entity.go
```diff
@@ -43,7 +48,13 @@ func newLocalEntity(storer persist.Storer, addHostnameToMetadata bool) *Entity {
 }
 
 // newEntity creates a new remote-entity.
-func newEntity(name, namespace string, storer persist.Storer, addHostnameToMetadata bool) (*Entity, error) {
+func newEntity(
+	name,
+	namespace string,
+	storer persist.Storer,
+	addHostnameToMetadata bool,
+	opts ...entityOption,
```
varas: I won't use func options here because:
- this won't ease customer usage as it's private
- this won't ease sdk-dev usage as there are no defined funcs, so you need to follow the whole flow to understand what's being done
- it's not consistent with rest of the options, either all func-opts, either none

integration/entity.go
```diff
@@ -17,6 +17,8 @@ type Entity struct {
 	Inventory   *inventory.Inventory `json:"inventory"`
 	Events      []*event.Event       `json:"events"`
 	AddHostname bool                 `json:"add_hostname,omitempty"` // add hostname to metadata at agent level
+	Cluster     string               `json:"cluster,omitempty"`      // add cluster to metadata at agent level
+	Service     string               `json:"service,omitempty"`      // add service to metadata at agent level
```
varas: we'll move these 2 fields into Metadata

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - SSL support for jmx package. Explained at https://github.com/newrelic/infra-integrations-sdk#jmx-support
 
+## Next Release
+
+### Changed
+
+- Protocol v2.1. NriCluster and NriService are set as entity attributes,
```
varas: v3

alejandrodnm: fixed

varas:
We'll replace this with identifier attributes.

PR #170, opened 2019-03-01
Protocol v3 doc

varas:

Documentation for

- [x] Protocol v3
- [x] Entity naming
- [x] New identifier attributes
- [ ] Update Agent dependency version in documentation.

### Author

### Reviewer

### DIFF
--- a/docs/entity-definition.md
+++ b/docs/entity-definition.md
@@ -1,23 +1,23 @@
-## What is an entity?
+# What is an entity?
 
 We use the vague term entity because we want to support hosts, pods, load balancers, DBs, etc. in a generic way. In the previous SDK versions (v1 & v2) the entity was local and just one, the host. In later versions the host reporting is called local entity, and it's optional to add metrics to it. You could just use remote entities to attach metrics. An entity can have its own inventory (configuration/state) and report any kind of metrics about itself. Although we may define a new entity for each monitored thing, we may want to relate/group some of them within a parent (local) entity (ie: host it's running on).
  
 Given we wanted to offer support for remote entities,let’s start by defining what we mean by an entity. **Entity** is a general term and refers to is a specific thing we collect data about. We used this term generally because we want to monitor different services and will need to support different parts such as hosts, pods, load balancers, DBs, etc. In the previous SDK versions (v1 & v2) the entity was local and just one, the host. However, as the entity does not necessarily have to be a host, we have broadened our support to include remote entities.
  
-### What do we mean by remote entity?
+## What do we mean by remote entity?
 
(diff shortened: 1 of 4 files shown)


docs/entity-definition.md
```diff
@@ -1,29 +1,104 @@
...
+
+Entity uniqueness is provided by its **key**.
+
+An entity `key` is formed by its `namespace` (or *type*) and `name`. These fields are concatenated with `:` separator.
+
+> For instance: `integration.Entity("name", "ns")` would create an entity with a `key` value `ns:name`.   
+
+Entities are attached to a user account. Each integration is responsible for providing uniqueness for its entities.
+
+It's up to the integration developer to define an entity `name` that's **unique** to its user account.
```
ikavanaghNR: An entity name should map uniquely to the entity being monitored. Imagine a situation where you are monitoring two Kafka clusters, one meant for development and the other for production, where you have a topic with the same name on each. The entity needs to map uniquely to the development topic and the production topic so that there is no collision.

ikavanaghNR:
```suggestion
It's up to the integration developer to define an entity `name` that's **unique** to its user account. An entity name should map uniquely to the entity being monitored. Imagine a situation where you are monitoring two Kafka clusters, one meant for development and the other for production, where you have a topic with the same name on each. The entity needs to map uniquely to the development topic and the production topic so that there is no collision.
```

docs/entity-definition.md
```diff
@@ -1,29 +1,106 @@
...
+
+> Eg: `host:port`
+
+
+### Entity name composition
+
+Names could be composed of several fields, in this case we call these **identifier-attributes**.
 
-### Metadata decoration
+#### Identifier attributes
```
ferranorriols: maybe it's quite obvious, but should we say the identifier attribute key should not use any special character like `:` or `=`?

ferranorriols:
Just adding these TODO tasks in order to keep in mind:

- [ ] update Release versions in changelog
- [ ] update Agent dependency version in documentation.

docs/entity-definition.md
varas: good point

ferranorriols: pefecgt

docs/protocol-v3.md
```diff
@@ -0,0 +1,34 @@
...
+
+- This localhost value does not provide valuable info without more context.
+- Name could collide with other service being named with a local address.
+
+This happens when:
+
+- Endpoints names are like `localhost:port`.
+- Ports tend be the same for a given service. Ie: 3306 for Mysql.
+
+
```
alejandrodnm: I would add a paragraph explaining what the agent does, something in the lines of:

> In this case, when the agent receives an entity key from a protocol v3 payload containing a loopback reference, it will replace such reference with ...

PR #172, opened 2019-03-04
CR for identifier attributes

varas:

Refactor: readability

### Author

### Reviewer

### DIFF
--- a/integration/entity.go
+++ b/integration/entity.go
@@ -25,8 +25,9 @@ type Entity struct {
 
 // EntityMetadata stores entity Metadata
 type EntityMetadata struct {
-	Name      string `json:"name"`
-	Namespace string `json:"type"` // For compatibility reasons we keep the type.
+	Name                 string              `json:"name"`
+	Namespace            string              `json:"type"`          // For compatibility reasons we keep the type.
+	IdentifierAttributes []map[string]string `json:"id_attributes"` // For entity key uniqueness
 }
 
 // newLocalEntity creates unique default entity without identifier (name & type)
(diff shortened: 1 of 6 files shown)


integration/entity.go
```diff
@@ -25,11 +25,14 @@ type Entity struct {
 
 // EntityMetadata stores entity Metadata
 type EntityMetadata struct {
-	Name                 string              `json:"name"`
-	Namespace            string              `json:"type"`          // For compatibility reasons we keep the type.
-	IdentifierAttributes []map[string]string `json:"id_attributes"` // For entity key uniqueness
+	Name      string   `json:"name"`
+	Namespace string   `json:"type"`          // For compatibility reasons we keep the type.
+	IDAttrs   []AttrKV `json:"id_attributes"` // For entity key uniqueness
```
ferranorriols: I prefer not using a customized type for readability reasons. Using `map[string]string` avoids the knowledge about `AttrKV`. I would go for new type like `AttrKV` for these cases:
- there is a clear method we want to attach to this new type
- we need the type for differentiate (a good example would be when we use values inside `context.Context`

integration/entity.go
```diff
@@ -67,18 +70,16 @@ func newEntity(
...
 	}
 
-	mapAttr := make([]map[string]string, len(idAttr))
-	for i, attr := range idAttr {
-		mapAttr[i] = map[string]string{
-			attr.key: attr.value,
-		}
+	idAttrKVs := []AttrKV{}
+	for _, a := range idAttrs {
+		idAttrKVs = append(idAttrKVs, a.KV())
```
ferranorriols: I prefer avoiding append here since we know the total length of the new slice. Using `make(x, len(idAttrKVs)` and then access by index we avoid new allocation in the slice.

integration/integration.go
```diff
@@ -49,6 +49,19 @@ type IdentifierAttribute struct {
 	value string
 }
 
+// NewIDAttr creates new identifier attribute.
+func NewIDAttr(key, value string) IdentifierAttribute {
```
ferranorriols: Since this `new` method is only used in tests and does not avoid a lot of complexity I would go for creating `map[string][string]`. Less is more ;-P

PR #178, opened 2019-04-09
EntityKey + attributes: ReportingEntityKey & ReportingEndpoint

varas:

We may also want to differentiate:

- The host/entity **producing** a remote entity data
- The host/entity **reporting** a remote entity data 

This gets relevance whenever the host reporting remote entity data is not the one producing it.

Usually for clusterised services a host/node entity reports data about the cluster entity. In this case we may want 
to know which entity (host/node) is reporting this cluster-entity data.

To know which entity is reporting current entity data we introduce the `integration.EntityKey` type and new attributes for metrics `reportingEntityKey` and `ReportingEndpoint`.

### Author

### Reviewer


docs/entity-definition.md
```diff
@@ -1,20 +1,55 @@
...
+
+**Entity** is a general term and refers to is a specific thing we collect data about. We used this term generally 
+because we want to monitor different services and will need to support different parts such as hosts, pods, load 
+balancers, DBs, etc. 
+
+
+## Local vs Remote entities
+
+In the previous SDK versions (v1 & v2) the entity was local and just one, the host. In later versions the host 
+reporting/forwarding data is called **local entity**, and it represent the host running the agent.
```
alejandrodnm: Shouldn't we try to deprecate **local entities**? Is there any scenario where they are still useful?

varas: Probably, but that's not the purpose of this PR

alejandrodnm: ok

PR #179, opened 2019-04-25
Entity fetching uses identifier attributes

varas:

[Identifier attributes](/blob/master/docs/entity-definition.md#identifier-attributes) are a coupled to the entity uniqueness. So different id-attrs should fetch/create different entities.

### Author

### Reviewer

### DIFF
--- a/integration/entity.go
+++ b/integration/entity.go
@@ -30,6 +30,26 @@ type EntityMetadata struct {
 	IDAttrs   IDAttributes `json:"id_attributes"` // For entity Key uniqueness
 }
 
+// EqualsTo returns true when both metadata are equal.
+func (m *EntityMetadata) EqualsTo(b *EntityMetadata) bool {
+	// prevent checking on Key() for performance
+	if m.Name != b.Name || m.Namespace != b.Namespace {
+		return false
+	}
+
+	k1, err := m.Key()
+	if err != nil {
+		return false
+	}
+
+	k2, err := b.Key()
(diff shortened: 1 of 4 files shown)


integration/entity.go
```diff
@@ -30,6 +30,26 @@ type EntityMetadata struct {
 	IDAttrs   IDAttributes `json:"id_attributes"` // For entity Key uniqueness
 }
 
+// EqualsTo returns true when both metadata are equal.
+func (m *EntityMetadata) EqualsTo(b *EntityMetadata) bool {
+	// prevent checking on Key() for performance
```
ferranorriols: I just wonder if really performance matters in this comparison. If I'm not wrong `integration.Entity` method is usually called at the beginning of the integration and it does not affect in performance.

In this case we could avoid all this `EqualsTo` & `sameAs` code and just use entity Key comparison here:
https://github.com/newrelic/infra-integrations-sdk/pull/179/files#diff-34fc2068cceaa93f8d6b312c44397541R159
```
e.Key() == eIt.Key()
```

varas: I think it matters as we don't know how integrations behave.

For instance an integration could be using `integration.Entity()` method to fetch entities. Instead of creating and keeping them in memory.

Also integration runtime is brought up and teared down on each integration run, so this method is going to be called frequently.

ferranorriols: fair enough

PR #180, opened 2019-05-28
Add positive versions of rate and delta

alejandrodnm:

This PR adds unsigned versions of RATE and DELTA, this is driven by the fact that we had a check for non-negative RATE/DELTA then remove it, and now it's being asked again. This adds the flexibility to use both cases.

https://github.com/newrelic/infra-integrations-sdk/pull/110

### Author

- [x ] add a risk label after carefully considering the "blast radius" of your changes
- [x ] describe the _intent_ of your changes in the description. don't just rewrite your code in prose
- [x ] assign at least one reviewer

### Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Adds the unsigned version of `RATE` and `DELTA`, `URATE` and `UDELTA`, which 
-  don't allow negatives values.
+- Adds the positive only version of `RATE` and `DELTA`, named `PRATE` and
+  `PDELTA` respectively.
 
 ## 3.1.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.2.0
+
+### Added
+
+- Adds the unsigned version of `RATE` and `DELTA`, `URATE` and `UDELTA`, which 
+  don't allow negatives values.
+
 ## 3.1.5
 
 ### Fixed
--- a/data/metric/metrics.go
+++ b/data/metric/metrics.go
@@ -28,6 +28,7 @@ var (
 	ErrNonNumeric        = errors.New("non-numeric value for rate/delta")
 	ErrNoStoreToCalcDiff = errors.New("cannot use deltas nor rates without persistent store")
 	ErrTooCloseSamples   = errors.New("samples too close in time, skipping")
+	ErrNegativeDiff      = errors.New("source was reset, skipping")
 	ErrOverrideSetAttrs  = errors.New("cannot overwrite metric-set attributes")
 	ErrDeltaWithNoAttrs  = errors.New("delta/rate metrics should be attached to an attribute identified metric-set")
 )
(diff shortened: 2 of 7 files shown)


data/metric/metrics.go
```diff
@@ -128,7 +130,7 @@ func castToFloat(value interface{}) (float64, error) {
 	return strconv.ParseFloat(fmt.Sprintf("%v", value), 64)
 }
 
-func (ms *Set) elapsedDifference(name string, absolute interface{}, sourceType SourceType) (elapsed float64, err error) {
+func (ms *Set) elapsedDifference(name string, absolute interface{}, sourceType SourceType, unsigned bool) (elapsed float64, err error) {
```
varas: I'd name this `unsiged` as `positive` as it's not converting to unsigned the elapsed diff, but returning errors on negative values

data/metric/metrics.go
```diff
@@ -80,12 +81,13 @@ func (ms *Set) SetMetric(name string, value interface{}, sourceType SourceType)
...
 	// Only sample metrics of numeric type
 	switch sourceType {
-	case RATE, DELTA:
+	case RATE, DELTA, URATE, UDELTA:
 		if len(ms.nsAttributes) == 0 {
 			err = ErrDeltaWithNoAttrs
 			return
 		}
-		newValue, errElapsed = ms.elapsedDifference(name, value, sourceType)
+		unsigned := sourceType == URATE || sourceType == UDELTA
```
varas: use method like `sourceType.IsUnsigned()` instead and you could inline the var

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.2.0
+
+### Added
+
+- Adds the unsigned version of `RATE` and `DELTA`, `URATE` and `UDELTA`, which
```
varas: shouldn't be `positive` instead of `unsigned` ?
I would expect an unsigned to return the elapsed diff with non signed value, not to assert that diff is positive

data/metric/metrics_test.go
```diff
@@ -94,6 +94,26 @@ func TestSet_SetMetricAllowsNegativeDeltas(t *testing.T) {
 	assert.Equal(t, ms.Metrics["d"], -3.0)
 }
 
+func TestSet_SetMetricUnsignedNotAllowsNegativeDeltas(t *testing.T) {
```
varas: add the success path to another test pls, maybe some of the ones above

CHANGELOG.md
alejandrodnm: I changed them to `PRATE` and `PDELTA`, I tried `POSTIVE_RATE` and `POSITIVE_DELTA` but the linter complained that it was not CamelCase, WDYT?

data/metric/source_type.go
```diff
@@ -48,6 +54,11 @@ func (t *SourceType) String() string {
 	return ""
 }
 
+// String fulfills stringer interface, returning empty string on invalid source types.
+func (t *SourceType) IsPositive() bool {
```
alejandrodnm: @varas I used a pointer receiver here because there was already a method with pointer receiver for the struct, I read some place that mixing pointer and value receivers for the same struct is a bad practice, the only ugly thing is that I had to reference the pointer to make the comparison.

Do you have any suggestion for this? or should I leave it like this?

CHANGELOG.md
varas: Linter is right consts should be camel-case unless they are borrowed from posix, but as we already have them in caps PDELTA lgtm

data/metric/metrics.go
```diff
@@ -163,6 +164,12 @@ func (ms *Set) elapsedDifference(name string, absolute interface{}, sourceType S
 	}
 
 	elapsed = newValue - oldValue
+
+	if elapsed < 0 && sourceType.IsPositive() {
```
ferranorriols: this is just a suggestion. Instead of having a `IsPositive` method that returns elapsed value and it validates depending the source, something like 
```
elapsed, err := sourceType.Elapsed(newValue, oldValue)

func (t SourceType) Elapsed(new, old int) (int, error) {
    e := n -old
    if e >= 0 {
        return e, nil
    }
    var err error
    if t. PRATE || t. PDELTA {
       err = ErrNegativeDiff
   }
   return e, err
}

```
In this case we delegate the responsability of validation to the source type and we don't have to check this in the `SetMetric` method.

Once again I like the current approach just to show how would be in this other way

PR #187, opened 2019-08-26
Enable custom attributes for events

alejandrodnm:

This PR allows events with attributes to be decorated with the entity `customAttributes` when they are added to via the `entity.AddEvent` function.

### Author

- [x ] add a risk label after carefully considering the "blast radius" of your changes
- [x ] describe the _intent_ of your changes in the description. don't just rewrite your code in prose
- [x ] assign at least one reviewer
- [x ] document feature

### Reviewer

### DIFF
--- a/data/attribute/attribute.go
+++ b/data/attribute/attribute.go
@@ -1,4 +1,4 @@
-package attributes
+package attribute
 
 import "fmt"
 
--- a/data/attributes/attributes.go
+++ b/data/attributes/attributes.go
@@ -1,0 +1,46 @@
+package attributes
+
+import "fmt"
+
+const (
+	// nsAttributeSeparator is the metric attribute key-value separator applied to generate the metric ns.
+	nsAttributeSeparator = "=="
+)
+
+// Attribute represents a metric attribute key-value pair.
+type Attribute struct {
+	Key   string
+	Value string
+}
+
+// Attributes list of attributes
+type Attributes []Attribute
+
+// Required for Go < v.18, as these do not include sort.Slice
+
+// Len ...
+func (a Attributes) Len() int { return len(a) }
+
+// Swap ...
+func (a Attributes) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+// Less ...
+func (a Attributes) Less(i, j int) bool {
+	if a[i].Key == a[j].Key {
+		return a[i].Value < a[j].Value
+	}
+	return a[i].Key < a[j].Key
+}
+
+// Namespace generates the string value of an attribute used to namespace a metric.
+func (a *Attribute) Namespace() string {
+	return fmt.Sprintf("%s%s%s", a.Key, nsAttributeSeparator, a.Value)
+}
+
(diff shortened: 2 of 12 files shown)


data/attributes/attributes.go
```diff
@@ -0,0 +1,46 @@
+package attributes
```
varas: singular for package names please

varas: is there any need to wrap this `Attribute` type into its own `attribute` package?

IMHO it feels more dev friendly to remain at the `metric` one for:
- atomicity: you won't use attributes without metrics
- simplicity: metrics package gets easier to use
- readability/usage: you don't need to type package name all around

integration/entity_test.go
```diff
@@ -103,6 +104,29 @@ func TestAddNotificationEvent(t *testing.T) {
 	}
 }
 
+func TestAddEventWithAttributes(t *testing.T) {
+	en, err := newEntity("Entity1", "Type1", persist.NewInMemoryStore(), false)
+	if err != nil {
+		t.Fatal(err)
```
varas: require.NoError

integration/entity_test.go
```diff
@@ -103,6 +104,29 @@ func TestAddNotificationEvent(t *testing.T) {
...
+	}
+
+	en.customAttributes = attributes.Attributes{attributes.Attr("clusterName", "my-cluster-name")}
+	attrs := map[string]interface{}{"attrKey": "attrVal"}
+	err = en.AddEvent(event.NewWithAttributes("TestSummary", "TestCategory", attrs))
+	assert.NoError(t, err)
+
+	assert.Len(t, en.Events, 1)
```
varas: require.Len

data/metric/metrics.go
```diff
@@ -59,20 +52,12 @@ func NewSet(eventType string, storer persist.Storer, attributes ...Attribute) (s
 }
 
 // AddCustomAttributes add customAttributes to MetricSet
-func AddCustomAttributes(metricSet *Set, customAttributes []Attribute) {
+func AddCustomAttributes(metricSet *Set, customAttributes []attributes.Attribute) {
```
varas: out of scope:
bad initial design, this should have been method and variadic, now modifying will cause BC

mariomac:
LGTM as long as Juan's comments are addressed

data/attribute/attribute.go
```diff
@@ -0,0 +1,46 @@
+package attribute
+
```
varas: https://github.com/newrelic/infra-integrations-sdk/pull/187#discussion_r326013176

data/attributes/attributes.go
alejandrodnm: Sorry I missed this. The reason to abstract it away into its own package is that now it used in both in the `metric` and `event` package. If you want I can keep it all on the `metric` but then the `event` package will have to import `metric` to use the attributes.

alejandrodnm:
@varas sorry i missed one of your comments, I replied here https://github.com/newrelic/infra-integrations-sdk/pull/187#discussion_r350369330

Issue #142, opened 2018-09-24
JMX Package Can Silently Drop Stderr When Command Fails


jaloren:
##### Description of the problem

I spent a couple hours on this but due to the package variables (global sadness) and the fine grained locking, I realized I would likely have ended up completely rewriting the package. So instead I'll simply highlight the problem and allow someone to make an incremental change.

The JMX command is run here with cmd.Wait:

https://github.com/newrelic/infra-integrations-sdk/blob/master/jmx/jmx.go#L99

If that fails, it sends on err channel in the goroutine. Open command will not return a non-nil error since that's in a separate goroutine. 

So then a doQuery func is run where it will send a string to os.Pipe, which since its a valid buffered pipe will succeed if the command is in the process of failing (still listening on the pipe) but has not yet exited. So I think there might be race condition.

https://github.com/newrelic/infra-integrations-sdk/blob/master/jmx/jmx.go#L135

So then we have receiveResult func, which has a select. The select is random. So the first case statement could evaluate and then you'll get "got empty result for query" error message so while there is data in stderr pipe its silently dropped.

https://github.com/newrelic/infra-integrations-sdk/blob/master/jmx/jmx.go#L179

##### Integrations SDK version
v2.1 

##### Golang version

go 1.10

##### OS

- Red Hat 7.5

varas:
Hi @jaloren, sorry for the late response. We'll release a fix for this issue soon.

About the global sadness within `jmx` package, you are def right. We know that it requires a whole rewrite but that's not high priority at the moment. 

cc @cristianciutea

varas:
Released at https://github.com/newrelic/infra-integrations-sdk/releases/tag/v3.5.0

PR #188, opened 2019-10-29
Fix JMX logging & error handling

varas:

Fix `jmx` package error handling & improve logging

*Disclaimer:*

`jmx` package is poorly designed and will be revamped in the future, this is trying to get it working with minimum effort investment.

### Author

- [ ] document feature

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -11,7 +11,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


varas:
Related issue https://github.com/newrelic/infra-integrations-sdk/issues/142

jmx/jmx.go
```diff
@@ -171,14 +169,35 @@ func openConnection(config *connectionConfig) error {
...
+			}
+		}
+	}()
+
 	if err = cmd.Start(); err != nil {
 		return err
 	}
 
 	go func() {
 		if err = cmd.Wait(); err != nil {
```
mariomac: I just faced this problem in the last week: you don't have to invoke cmd.Wait until you are sure that the standard error is closed, or cmd.Wait will actually close the standard error and you will loose error logs.

The approach for the above code should be something like:

```go
canCloseOutput := sync.WaitGroup{}
canCloseOutput.Add(1)
go func() {  // you could add a similar goroutine if you want to capture stdout this way
  defer canCloseOutput.Done()
  scanner := bufio.NewScanner(cmdError)
  scanner.Buffer([]byte{}, jmxLineBuffer)
  for scanner.Scan() {
    line := scanner.Text()
    if strings.HasPrefix(line, "WARNING") {
      cmdWarn <- fmt.Sprintf("cannot read nrjmx stderr: %s [proc-state: %s] (line: %s)", err, cmd.ProcessState, line)
    }
  }
}
var waitAndClose = func() {
  if err = cmd.Wait(); err != nil {
  if err != nil {
    cmdExitErr <- fmt.Errorf("nrjmx error: %s [proc-state: %s]", err, cmd.ProcessState)
  }
}
once := sync.Once{}
defer func() {
	canCloseOutput.Wait()
	once.Do(waitAndClose)
}()
// if the task is manually cancelled, waitAndClose is invoked explicitly
go func() {
	<-ctx.Done()
	once.Do(waitAndClose)
}()
```

varas: Nice one, but instead of having to close the pipe before calling Wait what is required is to drain the pipe channel: https://golang.org/pkg/os/exec/#Cmd.StderrPipe

mariomac: The above code does not close the pipe. It is automatically closed by the command when it has been drained. In the link you send:
> it is incorrect to call Wait before all reads from the pipe have completed. For the same reason, it is incorrect to use Run when using StderrPipe. See the StdoutPipe example for idiomatic usage.

But if you are only waiting for the stderr to be closed, maybe the above code can be simplified to:
```go
go func() {  // you could add a similar goroutine if you want to capture stdout this way
  scanner := bufio.NewScanner(cmdError)
  scanner.Buffer([]byte{}, jmxLineBuffer)
  for scanner.Scan() {
    line := scanner.Text()
    if strings.HasPrefix(line, "WARNING") {
      cmdWarn <- fmt.Sprintf("cannot read nrjmx stderr: %s [proc-state: %s] (line: %s)", err, cmd.ProcessState, line)
    }
  }
  if err = cmd.Wait(); err != nil {
    cmdExitErr <- fmt.Errorf("nrjmx error: %s [proc-state: %s]", err, cmd.ProcessState)
  }
}

varas: We cannot do wait on std reads as current API is so crappy there's global vars keeping the runtime context while execution is scattered btw `Open` (trigger cmd `Wait`) and `Query` reading the std outputs.

PR #194, opened 2019-11-05
JMX: custom nrjmx exec option + remove stderr closed

varas:

- JMX package  custom `nrjmx` executable tool option
- JMX package avoid nrjmx stderr closed error.

While reading from stderr, fd could have been closed by cmd.Wait, this is an expected error that doesn't cause any issue. We can safely ignore it.

> Going further this error comes from wrong jmx package API design, but this will be tackled in the future.

### Author

- [ ] document feature

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -34,7 +34,7 @@ var cmdExitErr = make(chan error, cmdStdChanLen)
 var done sync.WaitGroup
 
 var (
-	jmxCommand = "/usr/bin/nrjmx"
+	defaultNrjmxCommand = "/usr/bin/nrjmx"
 	// ErrJmxCmdRunning error returned when trying to Open and nrjmx command is still running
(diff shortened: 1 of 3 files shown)


jmx/jmx.go
```diff
@@ -63,7 +64,7 @@ func (cfg *connectionConfig) command() []string {
 	if os.Getenv("NR_JMX_TOOL") != "" {
 		c = strings.Split(os.Getenv("NR_JMX_TOOL"), " ")
```
davidbrota: why do we do the split here? Will this work with spaces inside the path? On windows, we install the tool on "c:\Program Files\New Relic\nrjmx\nrjmx.bat" which has some spaces.

varas: it works either way, I just tested it on Mac running a nrjmx tool with spaces in the patch and within the executable name itself

PR #203, opened 2019-11-13
Stop reading from stdOut after receiving message

ccheek21:
Previously, `doQuery` was only returning when an EOF was encountered.
However, this causes issues when multiple queries are to be sent to the
`nrjmx` process. Multiple `doQuery` goroutines were running at the
same time, and it appears the first one to run was always consuming
stdOut, which caused all remaining queries to fail with a timeout.

### Author

- [ ] document feature

### Reviewer

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -311,9 +311,7 @@ func doQuery(ctx context.Context, out chan []byte, queryErrC chan error, querySt
 			queryErrC <- fmt.Errorf("reading nrjmx stdout: %s", err.Error())
 		}
 		out <- b
-		if err != nil {
-			return
-		}
+		return
 	}
 }
 


ccheek21:
Tested with the following `nri-jmx` command. The beans it tries to collect should exist for any default JMX listener. 
```
bin/nr-jmx --collection_files /Users/ccheek/go/src/github.com/newrelic/nri-jmx/jvm-metrics.yml.sample --jmx_host tomcat-cent7-1 --jmx_port 9000 --jmx_user admin --jmx_pass admin --jmx_uri_path jmxrmi --timeout 5000 -verbose
```

This is with the following `jvm-metrics.yml.sample`
```
# Standard JVM Metrics

collect:
    - domain: java.lang
      event_type: JVMSample
      beans:
          - query: type=GarbageCollector,name=*
            attributes:
                - CollectionCount
                - CollectionTime
          - query: type=Memory
            attributes:
                - HeapMemoryUsage.Committed
                - HeapMemoryUsage.Init
                - HeapMemoryUsage.Max
                - HeapMemoryUsage.Used
                - NonHeapMemoryUsage.Committed
                - NonHeapMemoryUsage.Init
                - NonHeapMemoryUsage.Max
                - NonHeapMemoryUsage.Used
          - query: type=Threading
            attributes:
                - ThreadCount
                - TotalStartedThreadCount
          - query: type=ClassLoading
            attributes:
                - LoadedClassCount
          - query: type=Compilation
            attributes:
                - TotalCompilationTime
```

PR #208, opened 2020-03-27
use go modules

ardias:
Add support for go modules
Fixed linter issues with new more modern linter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,8 @@
 GO_VERSION = $(shell go version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
 
 GOTOOLS = github.com/axw/gocov/gocov \
-          github.com/AlekSi/gocov-xml \
+          github.com/AlekSi/gocov-xml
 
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -1,44 +1,62 @@
 GO_VERSION = $(shell go version | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
 
-GOTOOLS = gopkg.in/alecthomas/gometalinter.v2 \
-          github.com/axw/gocov/gocov \
+GOTOOLS = github.com/axw/gocov/gocov \
           github.com/AlekSi/gocov-xml \
 
 # golint only supports the last two Go versions, update the value when its not supported anymore.
 GOLINT_MIN_GO_VERSION = 1.9
```
davidbrota: not being used. should we add some version check?

Is the comment true? Latest two versions are 1.14 and 1.13

ardias: don't know if the comment is true. I'm guessing it was for previous versions. Will remove that

PR #212, opened 2020-04-02
first big api refactoring

ardias:
still needs some code cleanup
since I changes some of existing files, it may be hard to see the new api. 
see test **Test_Integration_PublishThrowsNoError** in integration_test.go for the full api


data/event/event.go
```diff
@@ -20,16 +23,18 @@ const (
...
+	Summary   string `json:"summary"`
+	Category  string `json:"category,omitempty"`
 	// Attributes are optional, they represent additional information that
 	// can be attached to an event.
 	Attributes map[string]interface{} `json:"attributes,omitempty"`
 }
 
 // New creates a new event.
-func New(summary, category string) *Event {
+func New(timestamp time.Time, summary, category string) *Event {
```
carlosroman, line 35: Why do we return a pointer rather than the object it self?

ardias: Just makes it a bit easier to use. Entity has as a slice of pointers to events, so you just need to "new"-it up and add it to the entity, no need to remember to add the & symbol.
We could change the AddEvent method to accept an Event instead of a pointer, but then you copyed the structure twice already, once from the New and another when adding it to the Entity, right?

carlosroman: Maybe `Entity` shouldn't be using a slice of pointers? Feels like a pre-optimisation because someone heard about the cost of copying structs around. Would be interesting to validate this as some tests have show pointers in slices can be worse.

ardias: may be. It's just seems to be the way most people do it. But it's true that makes it safer to use copying

ardias: i'll try to add benchmark tests

data/event/event.go
```diff
@@ -38,23 +43,16 @@ func New(summary, category string) *Event {
...
-	e := New(summary, category)
-	e.Attributes = attributes
-	return e
+// AddAttribute adds an attribute to the Event
+func (e *Event) AddAttribute(key string, value interface{}) {
+	// TODO validate value type (bool, number, string)
+	e.setAttribute(key, value)
 }
 
 func (e *Event) setAttribute(key string, val interface{}) {
```
davidbrota, line 55: we could merge this one with AddAttribute

data/event/event.go
davidbrota: the summary cannot be empty -> https://github.com/newrelic/infra-integrations-sdk/blob/b4b2f4383e084592c9
980f47644f5fda0e957280/integration/entity.go#L129

should we return an error here instead of there? to me it doesn't make any sense to create an Event which is not valid. In this case returning a nil would be better.

data/metadata/metadata.go
```diff
@@ -2,12 +2,26 @@ package metadata
...
+	Tags        TagMap `json:"tags"`
+}
+
+// New creates a new metadata section that "identifies" an entity
+func New(name string, entityType string, displayName string) *Metadata {
+	return &Metadata{
+		Name:        name,
+		DisplayName: displayName,
+		EntityType:  entityType,
+		Tags:        map[string]interface{}{},
```
davidbrota, line 23: Tags:        TagMap{} ?

data/metric/metrics.go
```diff
@@ -1,203 +1,110 @@
...
+			Dimensions: Dimensions{},
+		},
+		Value: value,
 	}
+}
 
-	if sourceType == RATE {
-		elapsed = elapsed / float64(duration)
+// NewCount creates a new metric of type Count
+func NewCount(timestamp time.Time, interval time.Duration, name string, count int64) Metric {
```
davidbrota, line 65: count can only be positive. Should we use a uint64 or return an error if the value is negative?

data/metric/metrics.go
```diff
@@ -1,203 +1,110 @@
...
-	separator := ""
-
-	attrs := ms.nsAttributes
-	sort.Sort(attribute.Attributes(attrs))
-
-	for _, attr := range attrs {
-		ns = fmt.Sprintf("%s%s%s", ns, separator, attr.Namespace())
-		separator = nsSeparator
+// NewSummary creates a new metric of type Summary
+func NewSummary(timestamp time.Time, interval time.Duration, name string, count int64, average float64, sum float64,
```
davidbrota, line 79: same here with this count parameter.

data/metric/source_type.go
```diff
@@ -14,7 +14,9 @@ type SourceType int
 // If any more SourceTypes are added update maps: SourcesTypeToName & SourcesNameToType.
 const (
 	// GAUGE is a value that may increase and decrease. It is stored as-is.
-	GAUGE SourceType = iota
+	GAUGE   SourceType = iota
+	COUNT   SourceType = iota
+	SUMMARY SourceType = iota
 	// RATE is an ever-growing value which might be reset. The package calculates the change rate.
 	RATE SourceType = iota
```
davidbrota, line 21: you can remove old types but pdelta.

integration/entity.go
```diff
@@ -16,7 +15,7 @@ import (
 type Entity struct {
 	Common    *Common              `json:"common"`
```
davidbrota, line 16: we are not going to support the common section for now.

data/metric/metrics.go
ardias: better a uint64. At first I had uint64 but then changed for some reason :)

integration/integration.go
```diff
@@ -248,3 +233,38 @@ func (i *Integration) checkArguments() error {
 
 	return errors.New("arguments must be a pointer to a struct (or nil)")
 }
+
+func (i *Integration) getAnonymousEntity() *Entity {
```
davidbrota, line 237: instead of this approach, you could add the anonymous entity inside the integration 
type Integration struct {
anonEntity Entity,
....
}
and then you add the metrics, events and inventory to that entity. Before creating the json, you can append that anonymous entity to the entity list (and remove it after the creating the json)

data/metric/source_type.go
ardias: i was going to ask which ones are we still supporting. 
pdelta means we have to "store" the metric and do the delta calculation? (i removed all that code)

integration/integration.go
ardias: good idea. probably safer to do that way

data/event/event.go
ardias: makes more sense to do it in the constructor. this should not be needed here.

data/event/event.go
```diff
@@ -19,42 +25,44 @@ const (
...
 	// Attributes are optional, they represent additional information that
 	// can be attached to an event.
 	Attributes map[string]interface{} `json:"attributes,omitempty"`
 }
 
 // New creates a new event.
-func New(summary, category string) *Event {
+func New(timestamp time.Time, summary, category string) (*Event, error) {
+	if len(summary) == 0 {
+		return nil, errors.New("summary cannot be empty")
```
davidbrota, line 45: err.ErrParameterCannotBeEmpty("summary") ?

integration/integration.go
```diff
@@ -40,12 +36,13 @@ type Integration struct {
 	ProtocolVersion    string    `json:"protocol_version"`
 	IntegrationVersion string    `json:"integration_version"`
 	Entities           []*Entity `json:"data"`
-	locker             sync.Locker
-	storer             persist.Storer
-	prettyOutput       bool
-	writer             io.Writer
-	logger             log.Logger
-	args               interface{}
+	// HostEntity is an "entity" that serves as dumping ground for metrics not associated with a specif entity
```
davidbrota, line 39: specif -> specific

integration/integration.go
```diff
@@ -248,3 +208,31 @@ func (i *Integration) checkArguments() error {
...
+				err := e.AddTag(strings.TrimPrefix(variable[0], prefix), variable[1])
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+
+	// TODO: should these custom "attributes" be added to "common", "metrics" and "events"?
+	if defaultArgs.NriCluster != "" {
```
davidbrota, line 230: NriCluster and NriService config options can be deleted

integration/integration.go
```diff
@@ -92,112 +89,51 @@ func New(name, version string, opts ...Option) (i *Integration, err error) {
...
+// Publish writes the data to output (stdout) and resets the integration "object"
 func (i *Integration) Publish() error {
-	if i.storer != nil {
-		if err := i.storer.Save(); err != nil {
-			return err
-		}
-	}
+	defer i.Clear()
 
+	// add the anon entity to the list of entity to be serialized, if not empty
```
davidbrota, line 127: anon -> host

PR #215, opened 2020-04-07
adias/add missing metric types

ardias:


.golangci.yml
```diff
@@ -18,6 +18,8 @@ linters-settings:
 
 issues:
   exclude-use-default: false
+  exclude:
+    - "don't use ALL_CAPS in Go names; use CamelCase"
```
davidbrota, line 22: 👏

data/metric/source_type.go
```diff
@@ -13,26 +13,41 @@ type SourceType int
 // Source types
 // If any more SourceTypes are added update maps: SourcesTypeToName & SourcesNameToType.
 const (
-	// GAUGE is a value that may increase and decrease. It is stored as-is.
+	// GAUGE is a value that may increase and decrease.
+	// It generally represents the value for something at a particular moment in time
 	GAUGE SourceType = iota
```
davidbrota, line 18: shouldn't we use CamelCase?

ardias: it was already in ALL_CAPS so i left it. It's not go "friendly", I can change that but for now we just ignoring the warning

PR #226, opened 2020-08-20
Edit to OSS files and README

theletterf:
Now that we have open sourced our instrumentation using the standardization best practices laid out in the open-by-default repository, we need to revisit repositories that were previously open source and make updates to achieve consistency in our open instrumentation repositories.


josemore (approved):
LGTM (apply same changes suggested in the infrastructure-agent repo)

PR #227, opened 2020-08-24
feat(metrics): allow NaN and Inf

ardias:
Nan or Inf values can be passed for summary metrics and will be serialized as null in the json output


carlosroman (approved):
Thanks for adding the test 👍

PR #235, opened 2020-09-02
adias/add prometheus histogram summary metrics

ardias:


gsanchezgavier (approved):
LGTM

data/metric/metrics.go
```diff
@@ -50,6 +50,35 @@ type summary struct {
 	Max     *float64 `json:"max"`
 }
 
+type bucket struct {
```
paologallinaharbur, line 53: Would make sense to use directly [the prometheus library](https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go) instead of our implementation?
 Or we are doing so not to be locked with such implementation/standard? In that case I would not call the struct `PrometheusHistogram` but simply `Histogram`

ardias: We could, but I wanted to avoid the dependency in here and in the agent. And this way we control the structure better.
We use the prefix Prometheus first because there is already a Summary metric from NR, and second because the agent needs to transform it and so if we use a "generic" name it could later conflict with other histogram implementation

paologallinaharbur: Fine for me! 👍

data/metric/metrics.go
```diff
@@ -177,6 +206,62 @@ func NewCumulativeRate(timestamp time.Time, name string, value float64) (Metric,
...
+func NewPrometheusHistogram(timestamp time.Time, name string, sampleCount uint64, sampleSum float64) (*PrometheusHistogram, error) {
+	return &PrometheusHistogram{
+		metricBase: metricBase{
+			Timestamp:  timestamp.Unix(),
+			Name:       name,
+			Type:       SourcesTypeToName[PROMETHEUS_HISTOGRAM],
+			Dimensions: Dimensions{},
+		},
+		// TODO make sure we make we don't allow NaN
+		SampleCount: &sampleCount,
```
paologallinaharbur, line 219: should we add a check also in the New function?

ardias: no need. this comment is obsolete

data/metric/source_type.go
```diff
@@ -28,26 +28,34 @@ const (
...
 
 // SourcesTypeToName metric sources list mapping its type to readable name.
 var SourcesTypeToName = map[SourceType]string{
-	GAUGE:            "gauge",
-	COUNT:            "count",
-	SUMMARY:          "summary",
-	CUMULATIVE_COUNT: "cumulative-count",
-	RATE:             "rate",
-	CUMULATIVE_RATE:  "cumulative-rate",
+	GAUGE:                "gauge",
```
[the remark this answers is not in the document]
ardias: done

carlosroman (approved):
LGTM, please raise issue on https://github.com/newrelic/infrastructure-agent for us to implement this change.

PR #237, opened 2020-09-04
feat(integration): add find entity by name

ardias:
This add support for finding an entity by name. This is required to avoid adding metrics belonging to the same entity in different entity entries (although they would end up being the same entity) and reducing the overall payload

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -176,6 +176,24 @@ func (i *Integration) Logger() log.Logger {
 	return i.logger
 }
 
+// FindEntity finds ad return an entity by name. returns false if entity does not exist in the integration
+func (i *Integration) FindEntity(name string) (*Entity, bool) {
+	if i.Entities == nil {
+		return &Entity{}, false
+	}
+	for _, e := range i.Entities {
+		// nil metadata usually means the "host" entity, so keep searching
+		if e == nil || e.Metadata == nil {
+			continue
+		}
+
+		if e.Metadata.Name == name {
+			return e, true
+		}
+	}
+	return &Entity{}, false
+}
+
 // Gauge creates a metric of type gauge
 func Gauge(timestamp time.Time, metricName string, value float64) (metric.Metric, error) {
 	return metric.NewGauge(timestamp, metricName, value)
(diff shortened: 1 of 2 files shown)


integration/integration.go
```diff
@@ -176,6 +176,20 @@ func (i *Integration) Logger() log.Logger {
 	return i.logger
 }
 
+// FindEntity finds ad return an entity by name. returns false if entity does not exist in the integration
+func (i *Integration) FindEntity(name string) (*Entity, bool) {
+	for _, e := range i.Entities {
+		if e.Metadata == nil {
+			return &Entity{}, false
```
gsanchezgavier, line 183: shouldn't be a continue here ? just to avoid checking a nil pointer in the following if

ardias: yes you're right we need a continue to keep searching..late friday code :)

varas (approved):

integration/integration.go
```diff
@@ -176,6 +176,21 @@ func (i *Integration) Logger() log.Logger {
 	return i.logger
 }
 
+// FindEntity finds ad return an entity by name. returns false if entity does not exist in the integration
+func (i *Integration) FindEntity(name string) (*Entity, bool) {
+	for _, e := range i.Entities {
+		// nil metadata usually means the "host" entity, so keep searching
+		if e.Metadata == nil {
```
paologallinaharbur, line 183: Entities is a slice of pointers... can `e` itself being nil or it is safe to take the assumption it will never happen?

Entities is exported so theoretically anyone can add an element to it

paologallinaharbur (approved):
Only a comment doubting if we need a extra if to check for a nil value or not

integration/integration.go
ardias: I think we can assume entities is not nil. When we create a new integration it sets entities to an empty slice. 
Someone can always set the slice to nil and although  don't think that's a realistic scenario might as well check for that

paologallinaharbur: no I mean `e` a generic element of the slice, not the slice itself.

ardias: Ah ok, i see what you mean.. It's a possibility, but then someone is doing the wrong thing almost on purpose..

ardias: > no I mean `e` a generic element of the slice, not the slice itself. Entities is defined as `Entities []*Entity` so elements can generate nullpointerExceptions

That's why it's a bad idea to have these things exported like this, it let's someone do the wrong things

PR #248, opened 2020-11-23
SDK v4

cristianciutea:


args/args.go
```diff
@@ -48,9 +47,9 @@ func (d *DefaultArgumentList) HasInventory() bool {
 // HTTPClientArgumentList are meant to be used as flags from a custom integrations. With this you could
 // send this arguments from the command line.
 type HTTPClientArgumentList struct {
-	HTTPCaBundleFile string        `default: "" help: "Name of the certificate file"`
-	HTTPCaBundleDir  string        `default: "" help: "Path where the certificate exists"`
-	HTTPTimeout      time.Duration `default:30 help: "Client http timeout in seconds"`
+	HTTPCaBundleFile string        `default:"" help:"Name of the certificate file"`
+	HTTPCaBundleDir  string        `default:"" help:"Path where the certificate exists"`
+	HTTPTimeout      time.Duration `default:"30" help:"Client http timeout in seconds"`
```
varas, line 52: shouldn't this be `"30s"`?

integration/integration.go
```diff
@@ -98,115 +96,51 @@ func New(name, version string, opts ...Option) (i *Integration, err error) {
...
 
+// AddEntity adds an entity to the list of entities. No check for "duplicates" is performed
+func (i *Integration) AddEntity(e *Entity) {
 	i.Entities = append(i.Entities, e)
+}
 
-	return e, nil
+// NewEvent creates a new event
+func (i *Integration) NewEvent(timestamp time.Time, summary string, category string) (*event.Event, error) {
+	return event.New(timestamp, summary, category)
```
varas, line 127: what's the goal?
I guess we should add the event to the integration, otherwise this seems to be just a helper function which I don't think does much besides increasing the API surface

integration/integration.go
```diff
@@ -229,26 +165,84 @@ func (i *Integration) MarshalJSON() (output []byte, err error) {
...
+
+// PrometheusSummary creates a metric of type prometheus summary
+func PrometheusSummary(timestamp time.Time, metricName string, sampleCount uint64, sampleSum float64) (*metric.PrometheusSummary, error) {
+	return metric.NewPrometheusSummary(timestamp, metricName, sampleCount, sampleSum)
+}
+
+// -- private
+// is entity empty?
+func notEmpty(entity *Entity) bool {
+	return len(entity.Events) > 0 || len(entity.Metrics) > 0 || len(entity.Inventory.Items()) > 0
```
varas, line 245: we might want a `Len()` method for Inventory instead

data/metadata/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Map stores the tags for the entity
+type Map map[string]interface{}
+
+// Metadata stores entity Metadata. Serialized as "entity"
+type Metadata struct {
+	// TODO: not a fan of having the fields exposed
+	Name        string `json:"name"`
+	DisplayName string `json:"displayName"`
+	EntityType  string `json:"type"`
```
varas, line 19: do we still want this? how this relates to UI shown entity-type?

data/metadata/metadata.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+const tagsPrefix = "tags."
+
+// Map stores the tags for the entity
+type Map map[string]interface{}
+
+// Metadata stores entity Metadata. Serialized as "entity"
+type Metadata struct {
+	// TODO: not a fan of having the fields exposed
```
varas, line 16: this isn't a TODO, just a complain

varas (commented):
I guess we can safely remove the `persist` package, right?

data/event/event.go
```diff
@@ -19,42 +24,44 @@ const (
...
-// AddCustomAttributes add customAttributes to the Event
-func AddCustomAttributes(e *Event, customAttributes []attribute.Attribute) {
-	for _, attr := range customAttributes {
-		e.setAttribute(attr.Key, attr.Value)
+// AddAttribute adds an attribute to the Event
+func (e *Event) AddAttribute(key string, value interface{}) error {
+	if len(key) == 0 {
+		return err.ParameterCannotBeEmpty("key")
 	}
+	// TODO validate value type (bool, number, string) ?
```
varas, line 64: ???

data/event/event.go
```diff
@@ -19,42 +24,44 @@ const (
...
-	e.Attributes[key] = val
+func NewNotification(summary string) (*Event, error) {
+	return New(time.Now(), summary, NotificationEventCategory)
 }
 
-// AddCustomAttributes add customAttributes to the Event
-func AddCustomAttributes(e *Event, customAttributes []attribute.Attribute) {
-	for _, attr := range customAttributes {
-		e.setAttribute(attr.Key, attr.Value)
+// AddAttribute adds an attribute to the Event
```
varas, line 59: how can one add attributes/fields to an event out of the `attributes` set?

infra agent converts these into `"attr.*"` keys, which isn't probably what users want/expect

maybe we should remove this prefix thing done by the agent?

varas (commented):
as we are validating telemetry data generation at SDK level we should import and use agent's `event. IsReserved()`
https://github.com/newrelic/infrastructure-agent/pull/224/files#diff-f35b8976e8af7ef3dd90969c4b2b48f8cdb23e77e2a79a2be4fea7225938e13dR40

args/args.go
cristianciutea: just saw that the sdk cannot parse time.Duration arguments, changing it to int

data/event/event.go
cristianciutea: I think we prefix them in the agent to avoid collision between collected data and custom attributes

data/event/event.go
cristianciutea: I don't think it's required, the agent handles this

data/metadata/metadata.go
cristianciutea: This is the field that the agent use on register as entityType

integration/integration.go
cristianciutea: fixed

integration/integration.go
cristianciutea: fixed

args/args.go
carlosroman: It should be able to parse `time.Duration` from a string into something like seconds.

PR #252, opened 2020-12-17
fix: Fixed incorrect use of go mod

carlosroman:


varas (approved):

data/event/event.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"time"
 
-	err "github.com/newrelic/infra-integrations-sdk/data/errors"
+	err "github.com/newrelic/infra-integrations-sdk/v4/data/errors"
 	agentEventPkg "github.com/newrelic/infrastructure-agent/pkg/event"
```
paologallinaharbur, line 8: @carlosroman Is it normal that the SDK depends on agent code?

PR #253, opened 2021-01-07
fix: warning stops mbeans processing

brushknight:
Problem: when SDK receives any warning from NRJMX integration, it stops to process remaining data. It has race condition inside, so sometimes it will process data first and exits, in some cases it will print warning first and exits.

Solution: make warning processing not terminating case (remove return from warning case) and add all warnings processing function to call as a defer for all return cases. This will help us with behaviour consistency

Chore: add SetOutput function for a logger, to be able to replace output stream for tests

### DIFF
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -341,7 +341,7 @@ func Query(objectPattern string, timeoutMillis int) (result map[string]interface
 
 // receiveResult checks for channels to receive result from nrjmx command.
 func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.CancelFunc, objectPattern string, timeout time.Duration) (result map[string]interface{}, err error) {
-	var warn string
+	defer logAvailableWarnings()
 	for {
 		select {
 		case line := <-lineC:
@@ -361,11 +361,7 @@ func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.Can
 			for k, v := range r {
 				result[k] = v
 			}
-			return
 
-		case warn = <-cmdWarnC:
-			// change on the API is required to return warnings
-			log.Warn(warn)
 			return
 
 		case err = <-cmdErrC:
@@ -379,6 +375,21 @@ func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.Can
 			cancelFn()
 			Close()
 			err = fmt.Errorf("timeout waiting for query: %s", objectPattern)
+			return
+		}
+	}
+}
+
+func logAvailableWarnings() {
+	var warn string
+	for {
+
+		select {
+		case warn = <-cmdWarnC:
+			{
+				log.Warn(warn)
+			}
+		default:
 			return
 		}
 	}
--- a/jmx/jmx.go
+++ b/jmx/jmx.go
@@ -390,9 +390,7 @@ func logAvailableWarnings(channel chan string) {
 	for {
 		select {
 		case warn = <-channel:
-			{
-				log.Warn(warn)
-			}
+			log.Warn(warn)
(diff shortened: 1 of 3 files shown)


jmx/jmx.go
```diff
@@ -361,11 +361,7 @@ func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.Can
 			for k, v := range r {
 				result[k] = v
 			}
-			return
 
-		case warn = <-cmdWarnC:
```
cristianciutea, line 366: Is there a reason for defering this? I'm afraid that as a side effect of consuming the channel later, it could lead to cmdWarnC being filled up and blocking potential errors being sent to error channel

brushknight: good point, we can increase queue, of restore this case

jmx/jmx.go
```diff
@@ -341,6 +341,7 @@ func Query(objectPattern string, timeoutMillis int) (result map[string]interface
 
 // receiveResult checks for channels to receive result from nrjmx command.
 func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.CancelFunc, objectPattern string, timeout time.Duration) (result map[string]interface{}, err error) {
+	defer logAvailableWarnings()
```
cristianciutea, line 344: do we still need this?

brushknight: yes, we want to print all warnings even if data was processes

brushknight: I moved global var as a param. We can create bool variable for the loop and move it after that, but it will increase complexity IMHO

jmx/jmx.go
```diff
@@ -383,3 +384,17 @@ func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.Can
 		}
 	}
 }
+
+func logAvailableWarnings(channel chan string) {
+	var warn string
+	for {
+		select {
+		case warn = <-channel:
+			{
```
varas, line 393: brackets not required here

jmx/jmx.go
```diff
@@ -383,3 +384,17 @@ func receiveResult(lineC chan []byte, queryErrC chan error, cancelFn context.Can
...
+				log.Warn(warn)
+			}
+		default:
```
gsanchezgavier, line 396: some time since i don't use the default clause in the select, is it not going to be executed if `channel` is empty? if so i think the for loop wrapping all this is not looping

jmx/jmx.go
```diff
@@ -323,6 +323,7 @@ func doQuery(ctx context.Context, out chan []byte, queryErrC chan error, querySt
 			queryErrC <- fmt.Errorf("reading nrjmx stdout: %s", err.Error())
 		}
 		out <- b
+		return
```
roobre, line 326: I do not understand the purpose of the `for` loop. Maybe I'm mistaken, but don't all execution paths lead to the loop breaking on the first iteration?

roobre:
Not a blocker for approval, but I see the code touched here is quite complex (even before your modifications). I would greatly appreciate if you could add some comments explaining what the functions do and when are they called now that you have it fresh, as it seems quite challenging to get right. Thanks!

jmx/jmx.go
brushknight: yes, you are right. As I understood, it was kept in hope to read all output from nrjmx, but it didn't happen

jmx/jmx.go
brushknight: We call this to clean warnings channel, if there is nothing, we are good to break the loop

varas (approved):

roobre (approved):

Issue #250, opened 2020-12-01
Setters and document the "common" attribute


gsanchezgavier:
There is a `common` [field ](https://github.com/newrelic/infra-integrations-sdk/blob/master/integration/entity.go#L15)in the payload that could be used for attributes shared by entities and metrics. This is not documented and also there are no methods instrumented to add this values. 

Would be good to document this attribute behavior with some examples, and also add the setters to this.

varas:
Released at https://github.com/newrelic/infra-integrations-sdk/releases/tag/v4.1.0

PR #254, opened 2021-01-15
feat: entity common dimensions support

varas:
Adding support for entity "common" dimensions. These ones are applied to every entity metric at platform level.

I'm not adding documentation for "common dimensions" as we [need to fix what a dimension is](https://github.com/newrelic/infra-integrations-sdk/issues/255).

Fixes https://github.com/newrelic/infra-integrations-sdk/issues/250


gsanchezgavier (approved):
Looks good!

PR #260, opened 2021-03-22
feat: add unique path for storage file

cristianciutea:
This PR will:
- add unique ID for integration instance based on integration cmd line arguments
- use the unique ID when creating the file for rate/delta storage (this will avoid collision issues when using discovery)
- add entityKey as a namespace attribute to the metric set

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -2,6 +2,7 @@ package integration
 
 import (
 	"bytes"
+	"crypto/md5"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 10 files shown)


gsanchezgavier (changes requested):
could you merge to v3 branch. I've already ported the jmx bug fix to that one

integration/integration.go
```diff
@@ -234,6 +241,14 @@ func (i *Integration) Logger() log.Logger {
 	return i.logger
 }
 
+// CreateUniqueID will generate an md5 string from integration arguments
+// to unique identify the integration instance.
+func (i *Integration) CreateUniqueID() string {
+	h := md5.New()
+	h.Write([]byte(fmt.Sprintf("%v", i.args)))
```
gsanchezgavier, line 248: are the configurations passed as environment variables inside args? I'm thinking of the conifguration sample for k8s that the integrations have it only uses env-vars to pass conifgurations and no cli args.

integration/integration_test.go
```diff
@@ -452,6 +452,38 @@ func TestIntegration_CreateLocalAndRemoteEntities(t *testing.T) {
...
+
+	var al argumentList
+
+	i, err := New("testIntegration", "0.0.0", Args(&al))
+	assert.NoError(t, err)
+
+	assert.Equal(t, i.CreateUniqueID(), "3071eb6863e28435e6c7e0c2bbe55ecd")
+}
+
+func TestIntegration_CreateUniqueID_EnvironmentVar(t *testing.T) {
```
gsanchezgavier, line 470: looks like this ut responds my question above :) nice

integration/integration.go
```diff
@@ -234,6 +241,14 @@ func (i *Integration) Logger() log.Logger {
 	return i.logger
 }
 
+// CreateUniqueID will generate an md5 string from integration arguments
+// to unique identify the integration instance.
+func (i *Integration) CreateUniqueID() string {
+	h := md5.New()
```
varas, line 247: not big deal, as integrations won't have enough instances to collide, but in case they did this may be a good option https://github.com/google/uuid

varas (approved):

PR #270, opened 2021-07-30
HTTP Client: Add possibility to skip server verification and refactor client to work with options pattern

alvarocabanas:
HTTP Client: Add possibility to skip server verification and refactor client to work with options pattern

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -14,12 +14,13 @@ import (
 	"time"
 )
 
-type clientOption func(*http.Client) error
+// ClientOption defines the format of the client option functions
+type ClientOption func(*http.Client) error
 
 // New creates a new http.Client with the Passed Client Options
 // that will have a custom certificate if it is loaded from the passed CA Bundle file and/or
 // directory. If both CABundleFile and CABundleDir are empty arguments, it creates an unsecure HTTP client.
-func New(opts ...clientOption) (*http.Client, error) {
+func New(opts ...ClientOption) (*http.Client, error) {
 	t := &http.Transport{
 		TLSClientConfig: &tls.Config{},
 	}
@@ -41,15 +42,17 @@ func New(opts ...clientOption) (*http.Client, error) {
 // if a CABundleDir is passed it will be joined to the CABundleFile to try to get the file.
 // If the file is a full path, even it the CABundleDir is passed, it will be detected and the full
 // path will be used instead
-func WithCABundleFile(CABundleFile, CABundleDir string) clientOption {
+func WithCABundleFile(CABundleFile, CABundleDir string) ClientOption {
 	return func(c *http.Client) error {
-		if CABundleFile != "" {
-			certPool := getClientCertPool(c)
-			if err := addCert(filepath.Join(CABundleDir, CABundleFile), certPool); err != nil {
-				if os.IsNotExist(err) {
-					if err = addCert(CABundleFile, certPool); err != nil {
-						return err
-					}
+		if CABundleFile == "" {
+			return nil
+		}
+
+		certPool := getClientCertPool(c)
+		if err := addCert(filepath.Join(CABundleDir, CABundleFile), certPool); err != nil {
+			if os.IsNotExist(err) {
+				if err = addCert(CABundleFile, certPool); err != nil {
+					return err
 				}
 			}
 		}
@@ -59,21 +62,23 @@ func WithCABundleFile(CABundleFile, CABundleDir string) clientOption {
 
 // WithCABundleDir adds the CABundleDir looks for pem certs in the
 // CABundleDir and adds them to the the client's certPool.
-func WithCABundleDir(CABundleDir string) clientOption {
+func WithCABundleDir(CABundleDir string) ClientOption {
 	return func(c *http.Client) error {
-		if CABundleDir != "" {
-			certPool := getClientCertPool(c)
-			files, err := ioutil.ReadDir(CABundleDir)
-			if err != nil {
-				return err
-			}
+		if CABundleDir == "" {
+			return nil
+		}
 
-			for _, f := range files {
-				if strings.Contains(f.Name(), ".pem") {
-					caCertFilePath := filepath.Join(CABundleDir, f.Name())
-					if err := addCert(caCertFilePath, certPool); err != nil {
-						return err
-					}
+		certPool := getClientCertPool(c)
+		files, err := ioutil.ReadDir(CABundleDir)
+		if err != nil {
+			return err
+		}
+
+		for _, f := range files {
+			if strings.Contains(f.Name(), ".pem") {
+				caCertFilePath := filepath.Join(CABundleDir, f.Name())
+				if err := addCert(caCertFilePath, certPool); err != nil {
+					return err
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


http/client.go
```diff
@@ -14,30 +14,45 @@ import (
...
 
-func _new(CABundleFile, CABundleDir string, httpTimeout time.Duration, acceptInvalidHostname string) (*http.Client, error) {
-	// go default http transport settings
-	t := &http.Transport{}
-
-	if CABundleFile != "" || CABundleDir != "" {
-		certs, err := getCertPool(CABundleFile, CABundleDir)
-		if err != nil {
-			return nil, err
+func WithCABundle(CABundleFile, CABundleDir string) TransportOption {
```
varas, line 38: this should be 2 different options, right?
I mean you setup the folder or the file, so either load the provided file or read all the certs from the folder

> there's the option to use the folder arg as a prefix path to the file, but I don't think this should be used at all

http/client.go
```diff
@@ -14,30 +14,45 @@ import (
 	"time"
 )
 
-// New creates a new http.Client with a custom certificate, which can be loaded from the passed CA Bundle file and/or
+type TransportOption func(*http.Transport) error
```
varas, line 17: good decision

varas (dismissed):
👏 good job!

http/client.go
```diff
@@ -14,30 +14,45 @@ import (
...
+			}
+			t.TLSClientConfig.RootCAs = certs
 		}
+		return nil
+	}
+}
 
-		t.TLSClientConfig = &tls.Config{
-			RootCAs: certs,
+func WithAcceptInvalidHostname(CABundleFile, CABundleDir, acceptInvalidHostname string) TransportOption {
```
roobre, line 51: I'm not 100% sure how `acceptInvalidHostname` relates to `CABundleFile` and `CABundleDir`. Is it a requirement for them to be tied together?

http/client.go
```diff
@@ -14,30 +14,45 @@ import (
...
-// New creates a new http.Client with a custom certificate, which can be loaded from the passed CA Bundle file and/or
+type TransportOption func(*http.Transport) error
+
+// New creates a new http.Client with the Passed Transport Options
+// that will have a custom certificate if it is loaded from the passed CA Bundle file and/or
 // directory. If both CABundleFile and CABundleDir are empty arguments, it creates an unsecure HTTP client.
-func New(CABundleFile, CABundleDir string, httpTimeout time.Duration) (*http.Client, error) {
-	return _new(CABundleFile, CABundleDir, httpTimeout, "")
-}
+func New(httpTimeout time.Duration, opts ...TransportOption) (*http.Client, error) {
```
roobre, line 22: Feels a bit strange to have the Options pattern but `httpTimeout` as a different parameter.

Unfortunately, timeout is not part of `http.Transport`... So we would need to change that from

`type TransportOption func(*http.Transport) error`
to
`type ClientOption func(*http.Client) error`

Which I'm not sure it's something we want. What are your thoughts? (also @varas)

http/client.go
alvarocabanas: I think it is, because if you check the VerifyOptions definition on the VerifyPeerCertificate, defines the Roots as the t.TLSClientConfig.RootCAs that are assigned on WithCABundle

http/client.go
alvarocabanas: I agree it's strange and thought the same, but if you check the actual logic, nothing it's avoiding you from setting both  at the same time. there is no return statement on the first `if certFile != "" { }` to stop executing the second block where the Dir is red, should I change it or it's a feature?

Anyway, even if we continues with the same logic where you can add both a file and a dir at the same time, I can separate WithCABundleDir and WithCABundleFile and pass them the x509.NewCertPool().  What do you think?

http/client.go
alvarocabanas: Since Timeout and Transport are public and can be modified after creation, I can definitelly do as you say, type ClientOption func(*http.Client), creating first of all the http.Client and modifying the timeout or the transport values in each configOption, do you agree?

http/client.go
roobre: +1 to the idea of splitting them. My understanding is that the code just checks for both options and appends them, so we better separate it to avoid uglyness and passing empty strings.

http/client.go
```diff
@@ -14,46 +14,63 @@ import (
...
+			}
+			c.Transport.(*http.Transport).TLSClientConfig.RootCAs = certs
+		}
+		return nil
+	}
+}
+
+func WithAcceptInvalidHostname(CABundleFile, CABundleDir, acceptInvalidHostname string) ClientOption {
+	return func(c *http.Client) error {
+		err := WithCABundle(CABundleFile, CABundleDir)(c)
```
roobre, line 55: We can probably "alias" the transport to avoid the verbosity

```suggestion
		transport := c.Transport.(*http.Transport)
		err := WithCABundle(CABundleFile, CABundleDir)(c)
```

http/client.go
roobre: If I got this correctly (which I'm not very sure about), this is a very convoluted way to convince the cert validation to check against a different hostname, instead of the one actually used for resolving the address.

It might be used together with custom CA bundles, but I don't see why it should be coupled. Do you have a different understanding?

http/client.go
```diff
@@ -14,46 +14,92 @@ import (
...
+	}
+
+	return client, nil
+}
+
+// WithCABundleFile adds the CABundleFile cert to the the client's certPool,
+// if a CABundleDir is passed it will be joined to the CABundleFile to try to get the file.
+// If the file is a full path, even it the CABundleDir is passed, it will be detected and the full
+// path will be used instead
+func WithCABundleFile(CABundleFile, CABundleDir string) ClientOption {
```
roobre, line 44: Not much a fan of this behavior, I'd prefer to just supply the path to the file directly.

http/client.go
```diff
@@ -14,46 +14,92 @@ import (
...
+			certPool := getClientCertPool(c)
+			if err := addCert(filepath.Join(CABundleDir, CABundleFile), certPool); err != nil {
+				if os.IsNotExist(err) {
+					if err = addCert(CABundleFile, certPool); err != nil {
+						return err
+					}
+				}
+			}
+		}
+		return nil
```
roobre, line 56: We could return an error `if CABundleFile == ""` and reduce indentation.

http/client.go
```diff
@@ -14,46 +14,92 @@ import (
...
+	}
+}
+
+// WithCABundleFile adds the CABundleDir looks for pem certs in the
+// CABundleDir and adds them to the the client's certPool.
+func WithCABundleDir(CABundleDir string) ClientOption {
+	return func(c *http.Client) error {
+		if CABundleDir != "" {
```
roobre, line 64: Same as above, I'd invert this check and return an error reducing indentation.

http/client.go
```diff
@@ -14,46 +14,92 @@ import (
...
 		}
+		return nil
+	}
+}
 
+// WithAcceptInvalidHostname allows the client to call the acceptInvalidHostname host
+// instead of the host from the certificates.
+func WithAcceptInvalidHostname(acceptInvalidHostname string) ClientOption {
+	return func(c *http.Client) error {
 		if acceptInvalidHostname != "" {
```
roobre, line 88: And same as the two above :)

http/client.go
varas, line 44: why don't we just remove the dir option and require file to be the path?

http/client.go
```diff
@@ -77,45 +123,33 @@ func _new(CABundleFile, CABundleDir string, httpTimeout time.Duration, acceptInv
...
 
-		for _, f := range files {
-			if strings.Contains(f.Name(), ".pem") {
-				caCertFilePath := filepath.Join(certDirectory, f.Name())
-				if err := addCert(caCertFilePath, caCertPool); err != nil {
-					return nil, err
-				}
-			}
-		}
+func getClientCertPool(c *http.Client) *x509.CertPool {
```
roobre, line 147: Could we get rid of the `get-` prefix? I don't think it adds much value.

roobre (changes requested):

http/client.go
```diff
@@ -14,46 +14,92 @@ import (
...
+		return nil
+	}
+}
 
+// WithAcceptInvalidHostname allows the client to call the acceptInvalidHostname host
+// instead of the host from the certificates.
+func WithAcceptInvalidHostname(acceptInvalidHostname string) ClientOption {
+	return func(c *http.Client) error {
 		if acceptInvalidHostname != "" {
+			transport := c.Transport.(*http.Transport)
```
varas, line 89: https://github.com/newrelic/infra-integrations-sdk/pull/270/files#r680792131

http/client.go
alvarocabanas: I know, but I don't want to break it for users who are using it, what do you think? this is a behaviour that is actually documented on the Unit Test, so it was a feature at the time.

alvarocabanas: I put it bellow, I agree, but we could be breaking the behaviour for some users.

varas: This is already a breaking change as we are modifying the constructor. So, we/sdk-users would notice about this anyway and would need to adjust it accordingly either way.

http/client.go
```diff
@@ -77,45 +123,33 @@ func _new(CABundleFile, CABundleDir string, httpTimeout time.Duration, acceptInv
...
-			if os.IsNotExist(err) {
-				if err = addCert(certFile, caCertPool); err != nil {
-					return nil, err
-				}
-			}
-		}
+// WithInsecureSkipVerify allows the client to call any host without checking the certificates.
+func WithInsecureSkipVerify() ClientOption {
+	return func(c *http.Client) error {
+		c.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify = true
```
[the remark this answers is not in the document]
alvarocabanas: Done

http/client.go
alvarocabanas: Done

http/client.go
alvarocabanas: Ok, modified!

alvarocabanas: Done

http/client.go
```diff
@@ -8,114 +8,165 @@ import (
...
 	"io/ioutil"
 	"net/http"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
 )
 
-// New creates a new http.Client with a custom certificate, which can be loaded from the passed CA Bundle file and/or
+var ErrInvalidTransportType = errors.New("roundTripper transport invalid, should be http type")
```
varas, line 16: missing docblock on exported stuff

http/client.go
```diff
@@ -8,114 +8,165 @@ import (
...
+				DNSName:       acceptInvalidHostname,
+				Intermediates: x509.NewCertPool(),
+				// On the server side, set KeyUsages to ExtKeyUsageClientAuth. The
+				// default value is appropriate for clients side verification.
+				// KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
 			}
+			for _, cert := range certs[1:] {
+				opts.Intermediates.AddCert(cert)
+			}
+			_, err := certs[0].Verify(opts)
```
varas, line 132: I know this wasn't touched, but shouldn't Verify be run for each cert?

varas (approved):

http/client.go
roobre: As far as I can tell this should be fine, as per the example here: https://github.com/golang/go/commit/29cfb4d3c3a97b6f426d1b899234da905be699aa

The first cert is the one we want to verify, and the rest are the issue chain, which get added into `opts`.

http/client.go
```diff
@@ -8,114 +8,166 @@ import (
...
 
-func _new(CABundleFile, CABundleDir string, httpTimeout time.Duration, acceptInvalidHostname string) (*http.Client, error) {
-	// go default http transport settings
-	t := &http.Transport{}
+// WithCABundleFile adds the CABundleFile cert to the the client's certPool
+func WithCABundleFile(CABundleFile string) ClientOption {
+	return func(c *http.Client) error {
+		if CABundleFile == "" {
+			return nil
+		}
```
roobre, line 48: Should we return an error in this case, rather than do nothing? Seems strange to invoke the option with an invalid argument.

varas: good point, OTOH there's this point about allowing functions to receive no-op argument

alvarocabanas: So you want to move the responsibility to the user of the sdk? ok, For sure I can do it.

alvarocabanas: Done

varas (dismissed):

http/client.go
```diff
@@ -8,114 +8,174 @@ import (
...
+	ErrInvalidTransportType = errors.New("roundTripper transport invalid, should be http type")
+	// ErrEmptyCABundleFile defines the error to be returned if the WithCABundleFile option is called with an empty value
+	ErrEmptyCABundleFile = errors.New("caBundleFile can't be empty")
+	// ErrEmptyCABundleDir defines the error to be returned if the WithCABundleDir option is called with an empty value
+	ErrEmptyCABundleDir = errors.New("caBundleDir can't be empty")
+	// ErrEmptyAcceptInvalidHostname defines the error to be returned if the WithAcceptInvalidHostname option is called with an empty value
+	ErrEmptyAcceptInvalidHostname = errors.New("acceptInvalidHostname can't be empty")
+)
+
+// ClientOption defines the format of the client option functions
```
roobre, line 27:
```suggestion
// ClientOption defines the format of the client option functions
// Note that many options defined in this package rely on the Transport for the http.Client being of type *http.Transport.
// Using a different Transport will cause these options to fail.
```

http/client.go
```diff
@@ -8,114 +8,174 @@ import (
...
-// New creates a new http.Client with a custom certificate, which can be loaded from the passed CA Bundle file and/or
+var (
+	// ErrInvalidTransportType defines the error to be returned if the transport defined for the client is incorrect
+	ErrInvalidTransportType = errors.New("roundTripper transport invalid, should be http type")
+	// ErrEmptyCABundleFile defines the error to be returned if the WithCABundleFile option is called with an empty value
+	ErrEmptyCABundleFile = errors.New("caBundleFile can't be empty")
+	// ErrEmptyCABundleDir defines the error to be returned if the WithCABundleDir option is called with an empty value
+	ErrEmptyCABundleDir = errors.New("caBundleDir can't be empty")
+	// ErrEmptyAcceptInvalidHostname defines the error to be returned if the WithAcceptInvalidHostname option is called with an empty value
+	ErrEmptyAcceptInvalidHostname = errors.New("acceptInvalidHostname can't be empty")
```
roobre, line 24: We could unify those into a single error:

`ErrEmptyArg = errors.New("argument is empty")`

And then decorate it on each function

```go
		if CABundleFile == "" {
			return fmt.Errorf("applying CABundleFile: %w", ErrEmptyArg)
		}
```

http/client.go
```diff
@@ -8,114 +8,174 @@ import (
...
+		return nil
 	}
+}
 
-	if certDirectory != "" {
-		files, err := ioutil.ReadDir(certDirectory)
-		if err != nil {
-			return nil, err
+// WithInsecureSkipVerify allows the client to call any host without checking the certificates.
+func WithInsecureSkipVerify() ClientOption {
```
roobre, line 149:
```suggestion
func WithTLSInsecureSkipVerify() ClientOption {
```

http/client.go
alvarocabanas: Done

http/client.go
```diff
@@ -148,7 +148,7 @@ func WithAcceptInvalidHostname(acceptInvalidHostname string) ClientOption {
 }
 
 // WithInsecureSkipVerify allows the client to call any host without checking the certificates.
```
varas: docblock first word should match the function name

roobre (approved):
Looking awesome, great job @alvarocabanas!

PR #272, opened 2021-08-02
jmx: wrap nrjmx errors so error details are not lost

roobre:
Motivated by the following line in the agent log:

`[ERR] Connection error for catalogone-kafka-prod3-east.vpc.verizon.com:9999 : jmx endpoint connection error\n`

Which doesn't really help to know the cause of the connection error.

If you agree with the change, I can port the commit to the `v3` branch as well.


rubenruizdegauna (approved):
LGTM

varas (approved):
👌

PR #286, opened 2022-01-27
fix: entity commondimension aligned with telemetry sdk

rogercoll:
Align common data with telemetry API: https://github.com/newrelic/newrelic-telemetry-sdk-go/blob/master/telemetry/metrics.go#L241

Infra agent: https://github.com/newrelic/infrastructure-agent/blob/30d0e84f3526ab0e1398d7908cc6e73193a21109/pkg/backend/telemetryapi/metrics_test.go#L52

### DIFF
--- a/integration/entity.go
+++ b/integration/entity.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/newrelic/infra-integrations-sdk/v4/data/event"
 	"github.com/newrelic/infra-integrations-sdk/v4/data/inventory"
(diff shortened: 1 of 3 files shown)


integration/entity.go
```diff
@@ -60,7 +70,25 @@ func (e *Entity) AddCommonDimension(key string, value string) {
...
+	t := timestamp.Unix()
+	e.CommonDimensions.Timestamp = &t
+}
+
+// AddCommonInterval adds a new common timestamp to the entity from duration in nanoseconds.
+func (e *Entity) AddCommonInterval(timestamp time.Duration) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+
+	t := timestamp.Milliseconds()
```
rubenruizdegauna, line 90: comment above says nanoseconds. Is this ok? ☝️

integration/entity_test.go
```diff
@@ -217,26 +217,35 @@ func Test_Entity_EntitiesWithSameMetadataAreSameAs(t *testing.T) {
 
 func TestEntity_AddCommonDimension(t *testing.T) {
 	tests := []struct {
-		name      string
-		commons   metric.Dimensions
-		expeceted *Entity
+		name     string
+		commons  metric.Dimensions
+		expected *Entity
```
rubenruizdegauna, line 222: what's wrong with `expeceted`??? :trollface:

rubenruizdegauna (approved):
LGTM! just a comment to clarify interval value

integration/entity_test.go
rogercoll: It is not familiar to me, YET 😂

integration/entity.go
rogercoll: The function parameter is in nanoseconds, the resulting interval in milliseconds.

New commit changed the comment as it was misleading. 

Thanks!

rubenruizdegauna: got it. Thanks!!

rubenruizdegauna (approved):
LGTM!

PR #288, opened 2022-02-18
feat: add host_id argument

brushknight:


integration/integration_test.go
```diff
@@ -463,6 +465,331 @@ func Test_Integration_PublishThrowsNoError(t *testing.T) {
...
+	_ = ev1.AddAttribute("attr1", "attr1Val")
+	_ = ev1.AddAttribute("attr2", 42)
+	e.AddEvent(ev1)
+	// add 2nd event to entity 1
+	ev2, err := event.New(time.Unix(10000000, 0), "evnt2sum", "evnt2cat")
+	assert.NoError(t, err)
+	e.AddEvent(ev2)
+	// add inventory to entity 1. only one because order is not guaranteed and the test is comparing with a static string
+	err = e.AddInventoryItem("custom/example", "version", "1.2.3")
+	assert.NoError(t, err)
```
gsanchezgavier, line 727: I would remove all this from the test and scope it just to test the `HostId` replacement logic. I think these should be checked in different unit tests.

brushknight: You are right, I will clean up this test a bit 👍

integration/integration.go
```diff
@@ -125,6 +130,13 @@ func (i *Integration) AddEntity(e *Entity) {
 func (i *Integration) Publish() error {
 	defer i.Clear()
 
+	hostID := i.GetHostID()
+	if hostID != "" {
+		for k := range i.Entities {
+			i.Entities[k].Metadata.Name = replaceLocalhost(i.Entities[k].Metadata.Name, hostID)
```
rubenruizdegauna, line 136: Do all entities always have metadata? If not we should check if it's not nil before accessing it. Maybe is not a common/real scenario but just in case.

rogercoll: All excepting the localhost one if they are created using the constructor https://github.com/newrelic/infra-integrations-sdk/blob/786cf26d3cdd3b8701c686be3f6a7ea25bc784b2/integration/integration.go#L105 

But as it is an exported struct I would also check it.

brushknight: added check from Metadata

gsanchezgavier (approved):
thanks

rubenruizdegauna (approved):
LGTM

rogercoll (approved):
lgtm

PR #289, opened 2022-03-02
Fix race condition new metric set

alvarocabanas:
Fix race condition when retrieving an entity at the same time as it's updated

### DIFF
--- a/integration/entity.go
+++ b/integration/entity.go
@@ -29,6 +29,7 @@ type EntityMetadata struct {
 	Name      string       `json:"name"`
 	Namespace string       `json:"type"`          // For compatibility reasons we keep the type.
 	IDAttrs   IDAttributes `json:"id_attributes"` // For entity Key uniqueness
+	lock      sync.Locker
 }
 
 // EqualsTo returns true when both metadata are equal.
@@ -89,6 +90,7 @@ func newEntity(
 			Name:      name,
 			Namespace: namespace,
 			IDAttrs:   idAttributes(idAttrs...),
+			lock:      &sync.Mutex{},
 		},
 	}
 
@@ -102,8 +104,6 @@ func (e *Entity) isLocalEntity() bool {
 
 // SameAs return true when is same entity
 func (e *Entity) SameAs(b *Entity) bool {
-	b.lock.Lock()
-	defer b.lock.Unlock()
 	if e.Metadata == nil || b.Metadata == nil {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


integration/entity.go
```diff
@@ -102,15 +102,18 @@ func (e *Entity) isLocalEntity() bool {
 
 // SameAs return true when is same entity
 func (e *Entity) SameAs(b *Entity) bool {
+	b.lock.Lock()
+	defer b.lock.Unlock()
 	if e.Metadata == nil || b.Metadata == nil {
```
rubenruizdegauna, line 107: out of curiosity, as `e.Metadata` is read here, shouldn't be also locked? Maybe I'm not totally getting the case but it  seems weird to lock `b` and not `e`

alvarocabanas: You are right, this responsibility shouldn't be delegated to the e entity. I've done it properly in the last commit, I've added a Mutex to the entityMetadata struct, that at the end is the one creating the race condition and I've locked the Key endpoint, and now everything works the same way but with a better design, thanks for pointing this out ;)

integration/entity.go
```diff
@@ -105,12 +107,13 @@ func (e *Entity) SameAs(b *Entity) bool {
...
 		return false
 	}
-
 	return e.Metadata.EqualsTo(b.Metadata)
 }
 
 // NewMetricSet returns a new instance of Set with its sample attached to the integration.
 func (e *Entity) NewMetricSet(eventType string, nameSpacingAttributes ...attribute.Attribute) *metric.Set {
+	e.lock.Lock()
+	defer e.lock.Unlock()
```
roobre, line 116: I think I'm lacking some context to understand the issue. Here we are moving the lock lines so we protect the modification of the metadata. when a new MetricSet is added. Then, we also lock the mutex when getting the entity Key. What problem does this solve?

alvarocabanas: It was not creating deadlock because it's two different mutexes, I basically created a new mutex for the entityMetadata and the one move upward is from the entity.
But anyway, moving this one upward was a inherited from my first solution where I was only using one mutex and now I removed the change moving it upward, and with the mutex in the key() method is enough to solve the race condition.

integration/entity.go
```diff
@@ -105,12 +107,13 @@ func (e *Entity) SameAs(b *Entity) bool {
...
 	s := metric.NewSet(eventType, e.storer, nameSpacingAttributes...)
 
 	if e.Metadata != nil {
```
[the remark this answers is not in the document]
alvarocabanas: If you refresh you'll see the change.

rubenruizdegauna (approved):
LGTM

roobre (approved):
LGTM

PR #290, opened 2022-03-08
POC Interface for adding metrics without entity

gsanchezgavier:
this is a quick dirty workaround to show what having an interface to add metrics not related to any entity would look like. 

IMO we should take the advantage that we are introducing some changes in the protocol to refactor the hole integration package to clearly show that the entity is not the main entry point to add metrics and even if possible i would get rid of most things related to entities since they will be synthesized in the backend.

### DIFF
--- a/integration/entity.go
+++ b/integration/entity.go
@@ -11,8 +11,8 @@ import (
 	"github.com/newrelic/infra-integrations-sdk/v4/data/metric"
 )
 
-// Entity is the producer of the data. Entity could be a host, a container, a pod, or whatever unit of meaning.
-type Entity struct {
+// DataSet is the producer of the data. DataSet could be a host, a container, a pod, or whatever unit of meaning.
+type DataSet struct {
 	CommonDimensions Common               `json:"common"` // dimensions common to every entity metric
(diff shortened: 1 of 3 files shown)


rogercoll:
Just to have a clear idea of the goal, is the following output a possible one after the refactor?

```json
{
  "protocol_version":"4",                      
  "integration":{                                                                         
    "name":"integration name",
    "version":"integration version"
  },
  "data":[                                   
    {
      "ignore_entity": true,
      "metrics":[                             
        {
          "name":"powerdns_something",
          "type":"count",                     
          "value":93, 
          "attributes":{}  
        }
      ],
      "common":{}                          
      "inventory":{},         
      "events":[]       
    },
    {
      "ignore_entity": false,
      "metrics":[                             
        {
          "name":"go_routines",
          "type":"count",                     
          "value":93, 
          "attributes":{}  
        }
      ],
      "common":{}                          
      "inventory":{},         
      "events":[]       
    },
  ]
}
```

Could we have multiple datasets from the same integration and that some datasets could be attached to the host entity and some not?

gsanchezgavier:
closing in favor of #291

Issue #284, opened 2022-01-25
Document SDK v4


rogercoll:
## Feature Description

Once https://github.com/newrelic/infra-integrations-sdk/issues/283 is implemented and the `Design Change: Entity synthesis for prometheus-based OHIs` has been tested and working, create a document with all the v4 fields, usage and consequences (data generated). 

- [ ] Host entity
- [ ] New entity
- [ ] Metrics/events
- [x] ignore_host_entity
- [ ] common, metadata, labels

Feel free to add more items into the list. 

The document should be uploaded in the repository and Confluence.

PR #291, opened 2022-03-09
Add ignore_entity attribute

rogercoll:


integration/entity.go
```diff
@@ -18,6 +18,7 @@ type Entity struct {
 	Metrics          metric.Metrics       `json:"metrics"`
 	Inventory        *inventory.Inventory `json:"inventory"`
 	Events           event.Events         `json:"events"`
+	IgnoreEntity     bool                 `json:"ignore_entity"`
```
gsanchezgavier, line 21: Could we add some documentations regarding this ?

integration/entity.go
```diff
@@ -114,6 +115,11 @@ func (e *Entity) AddMetadata(key string, value interface{}) error {
 	return nil
 }
 
+// UseHostEntity set if the host should use its entity
+func (e *Entity) UseHostEntity(value bool) {
```
gsanchezgavier, line 119: I would like to see a real an example (might be in the tutorial code) , my understanding is that we will need to do something like this 

```go
integration := New()
integration.HostEntity.UseHostEntity(false)
integration.HostEntity.AddMetric(myMetricNotRelatedToAnyEntity)
```
I would prefer the other hack creating an extra "Entity" already configured with this inside the `integration`

rogercoll: Sure, I will add a tutorial mentioning the entity synthesis workflow.

I really like the idea of the "Entity" already inside the integration but I might wait till the whole refactor as if we implement it right now there would be two confusing functions:

```go
integration.AddMetric()
entity.AddMetric()
```

Do you think it would be hard for POMI to change how those methods are called if we postpone the "hack" till the refactor?

rubenruizdegauna: I agree with adding doc/example as it is quite confusing right now. Also I would wait for the "hack" as there is no technical rush on it and this deserves a well thought refactor.

gsanchezgavier: > Do you think it would be hard for POMI to change how those methods are called if we postpone the "hack" till the refactor?

No, just confusing but we can document it there. 
I was hoping to define an interface that we would want to keep even after the refactor.

gsanchezgavier: I just document some of the key points of the refactor from the interface point of view here https://github.com/newrelic/infra-integrations-sdk/issues/292

gsanchezgavier: just the missing point of this thread would be to create an example or an integration test of how a user of this package can send metrics not associated to any entity.

cristianciutea (approved):

integration/entity.go
rogercoll: @gsanchezgavier I just added a brief summary into the v3tov4.md document. Nonetheless, the documentation needs a big refactor and we will approach that in: https://github.com/newrelic/infra-integrations-sdk/issues/284

PR #295, opened 2022-05-09
feat(persist): clean old entries upon loading store file

### LINKED ISSUE #294: [v3] metrics on the a storer file are not garbage collected
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )
There is no mechanism that removes metrics (after a TTL expired) from the [storer file](https://github.com/newrelic/infra-integrations-sdk/blob/v3/persist/storer.go#L66) used to save this metric into the disk to calculate the deltas on each execution. 

On initialization the integrations checks the TTL of the entire file remove it if it has expired. But there are cases where [the metric identifiers are based on ephemeral entities](https://github.com/newrelic/nri-varnish/issues/43) causing that new entries to this file are constantly being created and if the integration keeps it's execution normally the current clean up mechanism is never triggered, making this file grow without control and having the integration loading in on each execution.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )
Metrics on the storer should be garbage collected if they are not being updated after a TTL. 

Similar thing [is done in the Kubernetes integration ](https://github.com/newrelic/nri-kubernetes/blob/4d51d4928c28172032608b57c3c105039eda46af/internal/storer/storer.go#L108)

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )

## Your Environment
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )
One example happens on `nri-varnish` where `backend` entities could be ephemeral in some environments.
more context in this [issue](https://github.com/newrelic/nri-varnish/issues/43) 

## Additional context
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

gsanchezgavier:
Adds a garbage collect mechanism to remove entries from the storer file saved on disk which has been expired.
Expiration is calculated using the same `TTL`  that is being used to remove storer files and the mechanism remove the entires that has not been updated in a period greater than the `TTL`

Fixes #294 

**Context for the reviewer**:
- Storer files are created one per instance of the integration running on the host, since a hash, computed from the args, is [added to the file name](https://github.com/newrelic/infra-integrations-sdk/blob/v3/integration/integration.go#L103).
- Entries on the storer file are identified by the metric set attributes, usually entity attributes.
- The [current mechanism ](https://github.com/newrelic/infra-integrations-sdk/blob/v3/persist/store_path.go#L59)removes the hole file when it is not updated.
- The implemented mechanism remove entries that are not updated.

### DIFF
--- a/persist/storer.go
+++ b/persist/storer.go
@@ -23,10 +23,8 @@ const (
 	integrationsDir = "nr-integrations"
 )
 
-var (
-	// ErrNotFound defines an error that will be returned when trying to access a storage entry that can't be found
-	ErrNotFound = errors.New("key not found")
-)
+// ErrNotFound defines an error that will be returned when trying to access a storage entry that can't be found
+var ErrNotFound = errors.New("key not found")
 
 var now = time.Now
 
@@ -56,8 +54,12 @@ type inMemoryStore struct {
 
 // Holder for any entry in the JSON storage
 type jsonEntry struct {
+	entryTimestamp
+	Value interface{}
+}
+
+type entryTimestamp struct {
 	Timestamp int64
-	Value     interface{}
 }
 
 // fileStore is a Storer implementation that uses the file system as persistence backend, storing
@@ -122,6 +124,10 @@ func NewFileStore(storagePath string, ilog log.Logger, ttl time.Duration) (Store
 		if err != nil {
 			ilog.Debugf(err.Error())
 		}
+
+		if err := store.deleteOldEntries(ttl); err != nil {
+			ilog.Debugf("Error deleting old entries in store file %q: %w. Ignoring", store.path, err)
+		}
 	} else if os.IsNotExist(err) {
 		folder := path.Dir(storagePath)
 		err := os.MkdirAll(folder, dirFilePerm)
@@ -165,8 +171,8 @@ func (j inMemoryStore) Set(key string, value interface{}) int64 {
 
 	ts := now().Unix()
 	j.cachedData[key] = jsonEntry{
-		Timestamp: ts,
-		Value:     value,
+		entryTimestamp: entryTimestamp{ts},
+		Value:          value,
 	}
 	return ts
 }
@@ -235,6 +241,34 @@ func (j *fileStore) loadFromDisk() error {
 	if err != nil {
 		return fmt.Errorf("can't unmarshall %q: %s. Ignoring", j.path, err.Error())
 	}
+	return nil
+}
+
+func (j *fileStore) deleteOldEntries(ttl time.Duration) error {
+	j.locker.Lock()
+	defer j.locker.Unlock()
+
+	entry := entryTimestamp{}
+
+	unixTTL := int64(ttl.Seconds())
+	unixNow := time.Now().Unix()
+
+	var expiredKeys []string
+
+	for key, val := range j.Data {
+		if err := json.Unmarshal(val, &entry); err != nil {
+			return err
+		}
+
+		if (unixNow - entry.Timestamp) > unixTTL {
+			expiredKeys = append(expiredKeys, key)
+		}
+	}
+
+	for _, key := range expiredKeys {
+		delete(j.Data, key)
+	}
+
 	return nil
 }
 
--- a/persist/storer.go
+++ b/persist/storer.go
@@ -54,12 +54,8 @@ type inMemoryStore struct {
 
 // Holder for any entry in the JSON storage
 type jsonEntry struct {
-	entryTimestamp `json:",inline"`
-	Value          interface{}
-}
-
-type entryTimestamp struct {
 	Timestamp int64
+	Value     interface{}
 }
 
 // fileStore is a Storer implementation that uses the file system as persistence backend, storing
@@ -177,8 +173,8 @@ func (j inMemoryStore) Set(key string, value interface{}) int64 {
 
 	ts := now().Unix()
 	j.cachedData[key] = jsonEntry{
-		entryTimestamp: entryTimestamp{ts},
-		Value:          value,
+		Timestamp: ts,
+		Value:     value,
 	}
 	return ts
 }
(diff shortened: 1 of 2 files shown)


persist/storer.go
```diff
@@ -122,6 +124,10 @@ func NewFileStore(storagePath string, ilog log.Logger, ttl time.Duration) (Store
 		if err != nil {
 			ilog.Debugf(err.Error())
 		}
+
+		if err := store.deleteOldEntries(ttl); err != nil {
+			ilog.Debugf("Error deleting old entries in store file %q: %w. Ignoring", store.path, err)
+		}
```
roobre, line 130: After looking at the code for a while I now realize we are deleting old entries that might have been created _on a previous run_. Do you think we could add a comment for this? Otherwise readers might be confused to not see `deleteOldEntries` being called at the end of the lifecycle.

persist/storer.go
```diff
@@ -56,8 +54,12 @@ type inMemoryStore struct {
 
 // Holder for any entry in the JSON storage
 type jsonEntry struct {
+	entryTimestamp
```
roobre, line 57: If we are going to `json.Marshal` this, I think we need to
```suggestion
	entryTimestamp `json:",inline"`
```

Otherwise the go library will create an `entryTimestamp` node.

persist/storer.go
```diff
@@ -238,6 +244,34 @@ func (j *fileStore) loadFromDisk() error {
...
+	entry := entryTimestamp{}
+
+	unixTTL := int64(ttl.Seconds())
+	unixNow := time.Now().Unix()
+
+	var expiredKeys []string
+
+	for key, val := range j.Data {
+		if err := json.Unmarshal(val, &entry); err != nil {
+			return err
```
roobre, line 260: Naked error 🔞

persist/storer.go
```diff
@@ -238,6 +244,34 @@ func (j *fileStore) loadFromDisk() error {
...
+		}
+
+		if (unixNow - entry.Timestamp) > unixTTL {
+			expiredKeys = append(expiredKeys, key)
+		}
+	}
+
+	for _, key := range expiredKeys {
+		delete(j.Data, key)
+	}
```
roobre, line 270: Why do we store the keys to be deleted and then iterate over them, instead of deleting them in-place?

persist/storer_test.go
```diff
@@ -492,5 +474,60 @@ func TestFileStore_Save(t *testing.T) {
...
+
+	time.Sleep(time.Second * 2)
+	storer.Set("recentKey", "v")
+
+	assert.NoError(t, storer.Save())
+
+	storer, err = NewFileStore(filePath, log.NewStdErr(true), ttl)
+	assert.NoError(t, err)
+
+	time.Sleep(time.Second * 1)
```
roobre, line 502: Why are we sleeping an extra second here, after adding the entry we do not expect to be expired?

persist/storer.go
```diff
@@ -238,6 +244,34 @@ func (j *fileStore) loadFromDisk() error {
...
+	unixNow := time.Now().Unix()
+
+	var expiredKeys []string
+
+	for key, val := range j.Data {
+		if err := json.Unmarshal(val, &entry); err != nil {
+			return err
+		}
+
+		if (unixNow - entry.Timestamp) > unixTTL {
```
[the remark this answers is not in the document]
gsanchezgavier: I follow this approach to avoid any extra overhead since the timestamp is stored in `unix`. Perhaps i went too far  :)

persist/storer.go
gsanchezgavier: Oh i thought it was not safe but it actually is safe to do it 
https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

persist/storer_test.go
gsanchezgavier: oh this is from some tests i was doing in order to debug an issue was caused because i didn't set up `SetNow(time.Now)` . I will remove it thanks! it should be not needed

persist/storer.go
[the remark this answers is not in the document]
gsanchezgavier: I changed but still couldn't use the nice syntax since this package uses a custom `now()` and also to evaluate all entries with the same time.

persist/storer.go
gsanchezgavier: I actually re-think this and move it at the end on the save . This will give keys the chance to be updated in the current cycle.

persist/storer.go
gsanchezgavier: inline is the default behavior for embedded structs i think but i will add it to be more explicit , thanks!

persist/storer.go
```diff
@@ -56,8 +54,12 @@ type inMemoryStore struct {
...
 // Holder for any entry in the JSON storage
 type jsonEntry struct {
+	entryTimestamp `json:",inline"`
+	Value          interface{}
+}
+
+type entryTimestamp struct {
 	Timestamp int64
-	Value     interface{}
 }
```
rubenruizdegauna, line 64: out of curiosity, why this ☝️  change?

 If it's only for the unmarshall in `deleteOldEntries` I wouldn't change this struct but I would create an "anonymous" struct in `deleteOldEntries`

gsanchezgavier: > out of curiosity, why this ☝️ change?
Only timestamp is needed for the process of removing old entries. Decoding only this value saves some overhead (depending on the value stored), benchmark was almost half IIRC.

I started with that approach of anonymous and then moved here to have more transparent and hard dependency, but I don't have strong opinion on both, this just felt more safe to me in case of feature change WDYT?

gsanchezgavier: Looks good to me , I just changed

rubenruizdegauna (approved):
LGTM

PR #296, opened 2022-05-10
chore(ci): add test workflow

gsanchezgavier:
Adds same workflow of master (v4)


gsanchezgavier:
this is a bigger change that i expected, moving to an issue

PR #306, opened 2024-02-13
Add new TmpPath public function

alvarocabanas:
- Add new TmpPath public function to return a custom tmp path to the given integration's file.


rubenruizdegauna (dismissed):
LGTM

rubenruizdegauna (approved):

PR #315, opened 2024-09-19
Make timeout configurable (NR-281994)

kang-makes:
When users increase the interval to (for example) 5 minutes, our samples appear broken because the cache TTL is set by default to 1 minute making RATES and DELTAS to report only 0’s / nulls.

This PR set TTL as a variable and rise the default to 6 minutes (5 minutes plus a sane amount of time in case the integrations takes too long to execute)


integration/integration.go
```diff
@@ -89,6 +89,9 @@ func New(name, version string, opts ...Option) (i *Integration, err error) {
 	defaultArgs := args.GetDefaultArgs(i.args)
 	i.prettyOutput = defaultArgs.Pretty
 	i.addHostnameToMeta = defaultArgs.NriAddHostname
+	if defaultArgs.CacheTTL == 0 {
```
gsanchezgavier, line 92: IIRC the interval for the integration set in the config file is read by the infra-agent and then passed to the integrations through env vars , something like `NRI_INTERVAL`. Perhaps that could be used as default? so there is no need to set a TTL much larger than the interval?

gsanchezgavier: `NRI_CONFIG_INTERVAL`

kang-makes: There are (at least) two reasons to not do it:
* It would only work with that envvar from the agent and will misbehave if it is run standalone (by us, GTSEs or customers that know how to debug) giving weird metrics.
* It we want to tackled the issue I said above, an unclean default/fallback should be implemented.
* Cache refresh each time a key is accesses so a race condition can happen between the time the agent take to run the integration and the time that the integration takes to read that key. If the integration also takes more time to execute it might add more time to that run.

During the implementation of this feature, I wanted to do a breaking change to remove the default and make the storer to fail in case it is set to zero (leveraging that the integration create a storer instance and forcing the developer that is using the storer manualy to use it correctly)

I found many cases in the tests that made that refactor hard to do and started to became a huge change really fast.

This library has a lot of legacy and I would prefer to take the safe path for now.

sigilioso:
> This PR set TTL as a variable and rise the default to 6 minutes (5 minutes plus a sane amount of time in case the integrations takes too long to execute)

I think it is already planned but this change should be followed by an update in the docs to state what is the maximum supported interval value.

alvarocabanas (approved):
The changes look good and I have compiled nri-redis and made a simple test and everything works. 
I didn't test the specific behavior that is intended to tackle.

PR #316, opened 2024-09-20
Honor storer envvar (NR-302384)

kang-makes:
There a some integrations ([apache](https://github.com/newrelic/nri-apache/pull/132),  [cassandra](https://github.com/newrelic/nri-cassandra/pull/205), [nginx](https://github.com/newrelic/nri-nginx/pull/161), ...) that had a custom storer implementation.

This PR honors back that legacy behavior ins case there is any customer using that feature.

### DIFF
--- a/persist/storer.go
+++ b/persist/storer.go
@@ -95,6 +95,12 @@ func TmpPath(tempDir, integrationName string) string {
 // "nr-integrations" as a subfolder, Linux example: "/tmp/nr-integrations"
 func tmpIntegrationDir(tempDir string) string {
 	dir := tempDir
+
+	legacyCachePath := os.Getenv("NRIA_CACHE_PATH")
+	if legacyCachePath != "" {
+		tempDir = legacyCachePath
+	}
+
 	if tempDir == "" {
 		tempDir = os.TempDir()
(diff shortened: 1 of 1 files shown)


persist/storer.go
```diff
@@ -95,6 +95,12 @@ func TmpPath(tempDir, integrationName string) string {
 // "nr-integrations" as a subfolder, Linux example: "/tmp/nr-integrations"
 func tmpIntegrationDir(tempDir string) string {
 	dir := tempDir
+
+	legacyCachePath := os.Getenv("NRIA_CACHE_PATH")
```
gsanchezgavier, line 99: perhaps moving it one layer up, inside TmpPath? Ideally I would do it outside TmpPath but i guess we want to automatically get this for all integrations using the sdk.

gsanchezgavier (dismissed):
LGTM just left a nit, also if its possible to add some test would be nice

persist/storer.go
gsanchezgavier: make sense , i didn't noticed that other use

gsanchezgavier (approved):
lgtm

PR #317, opened 2024-10-02
Fix: go mod major version (NR-302384)

kang-makes:
Dependencies are not detecting this release as it does not have the major version in the package name.

This should fix the issue but it will force us to change all integrations to use the new package name.


alvarocabanas (approved):
LGTM

PR #320, opened 2026-05-29
fix: resolve CodeQL go/disabled-certificate-check

ujwalsah:
## Summary
- Replace `InsecureSkipVerify: true` + custom `VerifyPeerCertificate` with just `ServerName`
- Setting `ServerName` on the TLS config tells Go's default verification to check the certificate against that hostname instead of the connection address
- This achieves the same "accept invalid hostname" behavior while keeping full TLS verification enabled
- Resolves CodeQL alert `go/disabled-certificate-check` ([NR-560086](https://new-relic.atlassian.net/browse/NR-560086))

## Why this works
The original code set `InsecureSkipVerify: true` to disable Go's default hostname check, then re-implemented certificate chain verification in `VerifyPeerCertificate` with the overridden hostname. But `tls.Config.ServerName` already provides this — Go's default TLS verification uses `ServerName` for hostname matching when it's set, making the manual workaround unnecessary.

## Test plan
- [x] `go test ./http/...` passes
- [ ] Verify CI passes
- [ ] Confirm CodeQL alert auto-closes after merge

[NR-560086]: https://new-relic.atlassian.net/browse/NR-560086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


http/client.go
```diff
@@ -41,43 +41,11 @@ func _new(CABundleFile, CABundleDir string, httpTimeout time.Duration, acceptInv
 		}
 
 		if acceptInvalidHostname != "" {
```
shashank-reddy-nr, line 43: The suite only asserts success cases. There's no test that a hostname absent from the cert (or a bad CA) is rejected. Since the whole point of this PR is "verification is still enforced," a one-case negative test would lock that in and guard against future regressions:

```
// override hostname NOT present in the cert → should fail
client, _ := NewAcceptInvalidHostname(filepath.Join(tmpDir, "ca.pem"), "", time.Second, "wrong.host")
_, err = client.Do(req)
require.Error(t, err)
```

shashank-reddy-nr (approved):

shashank-reddy-nr (changes requested):
.

shashank-reddy-nr (approved):
LGTM

